Issue #98, opened 2023-05-04
Improve log field names


mck1117:
These names are all pretty bad for users, let's do something better:

![image](https://user-images.githubusercontent.com/568254/236313627-df42ea25-8cf8-4b7d-9acc-7c78ee6da770.png)

mck1117:
I have no idea how 035862e closed this, wat

PR #99, opened 2023-05-04
better data log names

nmschulte:
for #98


mck1117:
See how that file is in a folder called `generated`?

nmschulte:
hah ... been a long day.

PR #100, opened 2023-05-04
better output channel names

nmschulte:
for #98

### DIFF
--- a/firmware/console/binary/output_channels.txt
+++ b/firmware/console/binary/output_channels.txt
@@ -18,23 +18,23 @@ bit isO2HeaterOn;
 bit checkEngine;
 bit needBurn;
 
-bit sd_msd
-bit isFan2On
-bit acrActive;
-bit toothLogReady;
-bit isTpsError;
-bit isCltError;
-bit isMapError;
-bit isIatError;
-bit isTriggerError;
-bit hasCriticalError;
-bit isWarnNow;
-bit isPedalError;
-bit launchTriggered;
-bit isTps2Error;
-bit isIdleClosedLoop;
-bit injectorFault;
-bit ignitionFault;
+bit sd_msd;SD is MSD
+bit isFan2On;Radiator Fan2
+bit acrActive;Holley ACR Active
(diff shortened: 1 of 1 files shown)


firmware/console/binary/output_channels.txt
```diff
@@ -13,28 +13,28 @@ bit sd_present
...
-bit launchTriggered;
-bit isTps2Error;
-bit isIdleClosedLoop;
-bit injectorFault;
-bit ignitionFault;
+bit sd_msd;SD is MSD
+bit isFan2On;Radiator Fan2
+bit acrActive;Holley ACR Active
+bit toothLogReady;Tooth Logger Ready
+bit isTpsError;TPS Error
```
mck1117, line 25: How do we feel about something like `Error: TPS` so that these alphabetize better?

nmschulte: Sounds good to me.  Same for "Warning: ...", "Ready: ...", "Active: ..." ("Triggered: ..." -- change to "Active"?), "Fault: ..." (change to "Error"?)?

nmschulte:
force push fixes commit message typo ...

firmware/console/binary/output_channels.txt
mck1117: The other stuff doesn't seem to have many similar ones for different systems, let's just start with error flags for sensors and such

firmware/console/binary/output_channels.txt
```diff
@@ -13,28 +13,28 @@ bit sd_present
...
-bit isWarnNow;
-bit isPedalError;
-bit launchTriggered;
-bit isTps2Error;
-bit isIdleClosedLoop;
-bit injectorFault;
-bit ignitionFault;
+bit sd_msd;SD is MSD
+bit isFan2On;Radiator Fan2
+bit acrActive;Holley ACR Active
```
mck1117, line 23: Harley?

nmschulte: Of course.

PR #118, opened 2023-06-13
make license visible in main project page

mi-hol:


mck1117:
why does this require a rewrite of the whole file?

mi-hol:
Due to copyright terms in GPL3 doc
"Version 3, 29 June 2007

Copyright © 2007 Free Software Foundation, Inc. <<http://fsf.org/>>

Everyone is permitted to copy and distribute verbatim copies of this license
document, **but changing it is not allowed.**"

nmschulte:
Isn't the name `LICENSE` the typical place to find it?

Also, GNU provides the license in different formats: https://www.gnu.org/licenses/#GPL

mck1117:
supplemental terms are allowed, and do not constitute "changing" the license.

image

http://www.mmmtechlaw.com/wp-content/uploads/2014/03/PLI-Open-Source-2014-Extra-provisions-of-FSF-licenses.pdf

mi-hol:
good points, though my main intent was to "make license visible in main project page".
Any suggestions on: How can we achieve this goal **and** achieve your goals in combination?

mck1117:
> good points, though my main intent was to "make license visible in main project page".

can you test what happens if you move the "extra" stuff to the end of the license, instead of the beginning? then does github recognize it?

mi-hol:
> can you test what happens if you move the "extra" stuff to the end of the license, instead of the beginning? then does github recognize it?

It seems a bit more complex because:

1. GPL-3 license terms state:

   "If you add terms to a covered work in accord with this section, you must place, in
the relevant source files, a statement of the additional terms that apply to those
files, or a notice indicating where to find the applicable terms."

2. combined with the "no change of license" as [quoted above ](https://github.com/FOME-Tech/fome-fw/pull/118#issuecomment-1590073045) 

3.  a bit of reading in GH documentation about the license [parser](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#detecting-a-license)

Therefore it seems RE's approach to add terms directly into LICENSE.md is NOT compliant with GPL-3 terms

What is your conclusion?

mi-hol:
@mck1117 may I kindly ask for a reaction/comment?

nmschulte:
> good points, though my main intent was to "make license visible in main project page".

ah, I understand better now.

I see you closed this; what was the resolution?

mi-hol:
@nmschulte this is neither a reaction from Matt nor a solution.

PR #188, opened 2023-07-11
fix apparent CopyPaste error for assigned OBD fault codes

mi-hol:
4 error codes related to interpolation had been defined but only the first was used, therefore I assume a copy & paste with a missed change


mck1117:
why does this touch 1600 lines?

mi-hol:
> why does this touch 1600 lines?

Oops my fault, the 1600 lines change (commit text: "flag unused values clearly") was just a test of another issue related to a parser and should have gone to another branch.

Can I revert this commit via CLI (not so familiar with that)?

mi-hol:
> Can I revert this commit via CLI (not so familiar with that)?

`git revert HEAD~1 -m 1`  seems the correct command, right?

PR #204, opened 2023-07-20
fix link to CHANGELOG.rusefi.md

mi-hol:


mck1117:
why is it on a new line?

mi-hol:
I don't believe that links are allowed in headings

mck1117:
image

mi-hol:
ok I didn't know that but would still prefer to keep it as a separate line due to better usability.

- Readers don't expect links in headers 
- an md file converted to other format may have addl. restrictions

What reasons do you see to have just a header and no further content?

PR #206, opened 2023-07-21
remove livedata

mck1117:
it's just not useful compared to logs in TS


mck1117:
wow, this made the console launch really really snappy too

PR #215, opened 2023-07-26
smarter injector duty cycle limiting

mck1117:
Briefly allow >100% injector duty cycle. Includes supporting changes to make injector scheduling not fall apart at >100% duty.


firmware/controllers/limp_manager.cpp
```diff
@@ -148,11 +148,32 @@ void LimpManager::updateState(int rpm, efitick_t nowNt) {
 		allowFuel.clear(ClearReason::StopRequested);
 	}
 
-	// If duty cycle is high, impose a fuel cut rev limiter.
-	// This is safer than attempting to limp along with injectors or a pump that are out of flow.
-	// only reset once below 20% duty to force the driver to lift
-	if (m_injectorDutyCutHysteresis.test(getInjectorDutyCycle(rpm), 96, 20)) {
-		allowFuel.clear(ClearReason::InjectorDutyCycle);
+	{
+		// If duty cycle is high, impose a fuel cut rev limiter.
```
rusefillc, line 152: does scheduling code support 100% duty cycle?

mck1117: I believe it can, because of the pooled `scheduling_s` used for injection scheduling. Multiple injections get scheduled over top of one another, and the overlapping injection counters prevent closing the injector early.

I need to double check this on real hardware, though.

mck1117:
Seems to work, except there's a transient where it misses a single cycle at the instant it goes to 100% duty

image

mck1117:
Ooooooh look at that clean transition in/out of high duty operation!

image

PR #216, opened 2023-07-26
Update fome_console.xml to use open-source JAVA distribution

mi-hol:
I would strongly advise to switch to an open-source distribution due to ORACLE changing their license to be non-free some time ago. Recommended is a version with long-time support (LTS)

### DIFF
--- a/misc/console_launcher/fome_console.xml
+++ b/misc/console_launcher/fome_console.xml
@@ -8,7 +8,7 @@
   <cmdLine></cmdLine>
   <chdir>.</chdir>
   <priority>normal</priority>
-  <downloadUrl>http://java.com/download</downloadUrl>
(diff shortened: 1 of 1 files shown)


nmschulte:
Concur, though IIUC we're in the clear about the license (non-commercial).  But yes, OpenJDK is the way IMO.

mck1117:
I'll actually push back on this a little - there's potential for "ordinary users" to end up in a weird state here if they get recommended to install non-Oracle Java with it already installed.

This file has no control over what JRE the console uses, only what the launcher will recommend if you don't have it installed.

mi-hol:
> there's potential for "ordinary users" to end up in a weird state

Would you mind to elaborate?
I had multiple ORACLE and non-ORACLE JRE installed in the past, the only issue I remember was that "java.exe" used the version first found in PATH variable.

PR #223, opened 2023-08-11
comment unused custom FOME OBD codes

mi-hol:
clarifies the codes unused and potentially reduces size of binary for debug build (that hit limits occasionally)


mck1117:
> reduces size of binary for debug build (that hit limits occasionally)

what size limit?

mi-hol:
> what size limit

I seem to remember that RE debug builds failed but I may have got this wrong.
Nevertheless the main goal is to "cleanup" by flagging unused values (and there are many more)

mck1117:
Can we do a smaller step first of renaming unused options back to `CUSTOM_ERR_xxxx`? Not sure of the value of commenting out some small set of unused codes in the list.

mck1117:
A potentially more useful pursuit is giving names to the ones that are used, but are named with just a code. For example:

image

`CUSTOM_ERR_6566` could be something like `CUSTOM_ERR_CYL_ANGLE`.

mi-hol:
> Not sure of the value of commenting out some small set of unused codes in the list.

benefits are:
- allow for removal of unused codes in documentation (please see discord https://discord.com/channels/1060875162892898324/1100956878349017128/1126940464814231612)
- enables automatic updates of documentation from source code
- make it 100% clear what are the available error codes for programmes without search through code base first
=> efficiency gain

> A potentially more useful pursuit is giving names to the ones that are used, but are named with just a code.

Happy to tackle this after removing the unused codes as the next step

mi-hol:
> * enables automatic updates of documentation from source code

is now further explained in https://github.com/FOME-Tech/fome-fw/issues/227.
Please comment there

mi-hol:
> > A potentially more useful pursuit is giving names to the ones that are used, but are named with just a code.
> 
> Happy to tackle this after removing the unused codes as the next step.

Are we in a "deadlock" situation where both of us expect the other to make a move?

PR #225, opened 2023-08-14
Fix compiler error in simulator

ConnorRigby:
> ../firmware/console/binary_log/log_fields_generated.h:351:1: error: the value of ‘engine’ is not usable in a constant expression


mck1117:
why doesn't this fail in CI builds? Can we provoke this error in CI builds?

mck1117:
@ConnorRigby I've never seen this error before, where do you see it occur?

PR #230, opened 2023-08-28
enable f7 openblt

mck1117:


mck1117:
current status:
- Bootloader operates
- Can connect to it from PC
- Can flash firmware from PC
- Flashed image exactly matches expected file generated on PC by stitching together
- Bootloader checksum verification works
- Whether flashed via the combined image or via OpenBLT, the firmware doesn't boot :(

mck1117:
Seems to somehow depend on optimization settings of the main FW:

- O0 works
- Og works
- O1 and up doesn't work

mck1117:
it's making a wild jump in to nowhere land on this `isb` in `SCB_EnableDCache`

image

PR #238, opened 2023-09-11
fix apparent CopyPaste error when assigning obd fault codes

mi-hol:
was merged in RE already


mi-hol:
@mck1117 does reviewing this require more than 3 minutes?

mck1117:
> @mck1117 does reviewing this require more than 3 minutes?

Yes, it does. I don't understand why the description "fix copy paste error" requires changing 1600 lines of a file. It adds `unused code:` to many lines, but I think that was already obvious as those lines were commented out. But it also touches some other file's OBD codes, so there's some other edit of a few lines "hidden" in the middle of those 1600 lines.

mck1117:
looks like there are also numerous mis-edits

image

mi-hol:
getting crap in was easy, removing it is quite difficult

PR #247, opened 2023-09-16
add knock frequency power-cycle tooltip

### LINKED ISSUE #245: Knock Frequency tool tip
Add to the TS-tip "Changing this setting requires a power cycle". 

Low priority, just one I identified that always needs a reset and could catch people out.

nmschulte:
resolves #245


mck1117:
There's also a section in the .input of "powerCycleRequired" or something so that TS warns about it.

PR #273, opened 2023-10-26
switch openblt to FOME fork

mck1117:


mck1117:
Tested out OK with the proteus f4 build.

PR #275, opened 2023-10-28
Fixed FOME CAN data bus selection

Krakert:
pass the CAN bus via the `transmitStruct` to `CanTxMessage::CanTxMessage`, bus is by default 0, now looks at the setting from TS, `bool canBroadcastUseChannelTwo`.

### DIFF
--- a/firmware/controllers/can/can_dash.cpp
+++ b/firmware/controllers/can/can_dash.cpp
@@ -1293,14 +1293,16 @@ void canDashboardAim(CanCycle cycle) {
 		return;
 	}
 
-	transmitStruct<Aim5f0>(0x5f0, false);
-	transmitStruct<Aim5f1>(0x5f1, false);
-	transmitStruct<Aim5f2>(0x5f2, false);
-	transmitStruct<Aim5f3>(0x5f3, false);
-	transmitStruct<Aim5f4>(0x5f4, false);
-	transmitStruct<Aim5f5>(0x5f5, false);
-	transmitStruct<Aim5f6>(0x5f6, false);
-	transmitStruct<Aim5f7>(0x5f7, false);
+	auto canChannel = engineConfiguration->canBroadcastUseChannelTwo;
+
+	transmitStruct<Aim5f0>(0x5f0, false, canChannel);
+	transmitStruct<Aim5f1>(0x5f1, false, canChannel);
+	transmitStruct<Aim5f2>(0x5f2, false, canChannel);
+	transmitStruct<Aim5f3>(0x5f3, false, canChannel);
+	transmitStruct<Aim5f4>(0x5f4, false, canChannel);
+	transmitStruct<Aim5f5>(0x5f5, false, canChannel);
+	transmitStruct<Aim5f6>(0x5f6, false, canChannel);
+	transmitStruct<Aim5f7>(0x5f7, false, canChannel);
 
 	// there are more, but less important for us
 	// transmitStruct<Aim5f8>(0x5f8, false);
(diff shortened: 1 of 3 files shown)


firmware/controllers/can/can_verbose.cpp
```diff
@@ -216,17 +216,18 @@ static void populateFrame(Odometry& msg) {
 void sendCanVerbose() {
 	auto base = engineConfiguration->verboseCanBaseAddress;
 	auto isExt = engineConfiguration->rusefiVerbose29b;
-
```
nmschulte, line 219: omg the tabs! :laughing:

firmware/hw_layer/drivers/can/can_msg_tx.h
```diff
@@ -112,9 +112,9 @@ class CanTxTyped final : public CanTxMessage
 };
 
 template <typename TData>
-void transmitStruct(uint32_t id, bool isExtended)
+void transmitStruct(uint32_t id, bool isExtended, bool canChannel = false)
```
nmschulte, line 115: shouldn't this be a `size_t` ("`bus`")?

line 31:
```c++
explicit CanTxMessage(uint32_t eid, uint8_t dlc = 8, size_t bus = 0, bool isExtended = false);
```

mck1117: let's make this parameter `canChannel` non-optional - that's sort of what got us in this mess in the first place

mck1117: > so that inside the function we look at which bus to use?

eh? All I mean by my comment was to remove `= false` on line 115 here.

Krakert: When I change that, I will update all uses for `transmitStruct`

firmware/controllers/can/can_dash.cpp
```diff
@@ -1293,14 +1293,16 @@ void canDashboardAim(CanCycle cycle) {
...
 
-	transmitStruct<Aim5f0>(0x5f0, false);
-	transmitStruct<Aim5f1>(0x5f1, false);
-	transmitStruct<Aim5f2>(0x5f2, false);
-	transmitStruct<Aim5f3>(0x5f3, false);
-	transmitStruct<Aim5f4>(0x5f4, false);
-	transmitStruct<Aim5f5>(0x5f5, false);
-	transmitStruct<Aim5f6>(0x5f6, false);
-	transmitStruct<Aim5f7>(0x5f7, false);
+	auto canChannel = engineConfiguration->canBroadcastUseChannelTwo;
```
rusefillc, line 1296: Is not "canChannel" a less than ideal name for "use second channel"? also what if _soon_ more than two buses would be supported? A few of stm32f7 have three CAN buses.

firmware/hw_layer/drivers/can/can_msg_tx.h
```diff
@@ -91,7 +91,7 @@ class CanTxTyped final : public CanTxMessage
 #endif // EFI_CAN_SUPPORT
 
 public:
-	explicit CanTxTyped(uint32_t id, bool isExtended) : CanTxMessage(id, sizeof(TData), isExtended) { }
+	explicit CanTxTyped(uint32_t id, bool isExtended, bool canChannel) : CanTxMessage(id, sizeof(TData), canChannel, isExtended) { }
```
rusefillc, line 94: bool to size_t conversion heh?

rusefillc (approved):

firmware/controllers/can/can_dash.cpp
Krakert: Now `canBroadcastUseChannelTwo` is a Bool, CAN 2 is a `true / 1` so when more CAN buses are added we can switch to a int which keeps the logic the same.

rusefillc: So why not "size_t busIndex engineConfiguration->canBroadcastUseChannelTwo;" to keep the... technical debt most localized?

mck1117: yea, that was going to be my next comment, we should really be doing `size_t busIndex = engineConfiguration->canBroadcastUseChannelTwo ? 1 : 0`

PR #276, opened 2023-10-29
consistent line endings in log_fields_generated.h

nmschulte:


mck1117:
now I'm curious, do we have any other \r\n?

nmschulte:
> now I'm curious, do we have any other \r\n?

Yes, there are many.

PR #291, opened 2023-11-02
one thing per line is nicer for history etc

rusefillc:


firmware/rusefi_rules.mk
```diff
@@ -3,4 +3,9 @@ RUSEFI_OPT = -Werror
 RUSEFI_CPPOPT = -Werror=shadow
 
 # ...except these few
-RUSEFI_OPT += -Wno-error=unused-function -Wno-error=unused-variable -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=undef -Wno-error=stringop-overflow
+RUSEFI_OPT += -Wno-error=unused-function
```
nmschulte, line 6: I like this change.  also, a commonly used shorter syntax (but w/ last-line concerns):
```sh
RUSEFI_OPT += -Wno-error=unused-function \
    -Wno-error=unused-variable \
    -Wno-error=sign-compare
    -Wno-error=unused-parameter \
    -Wno-error=undef \
    -Wno-error=stringop-overflow
```

nmschulte (approved):

firmware/rusefi_rules.mk
mck1117: you can leave the `\` on the last line so long as there's an empty line after it 🙂

PR #301, opened 2023-11-09
unit tests on Windows

rusefillc:
asserting that Windows development is possible


mck1117:
Builds on windows are not supported by FOME, so tests are not required to work there.

rusefillc:
> Builds on windows are not supported by FOME, so tests are not required to work there.

https://github.com/FOME-Tech/fome-fw/pull/392

PR #312, opened 2023-12-06
Only primary needs cam resync counter

rusefillc:


rusefillc:
probably false red HW CI error?

mck1117:
why is generated checked in?

rusefillc:
@mck1117 try now

PR #316, opened 2023-12-13
Code duplication

rusefillc:


rusefillc:
@mck1117 HW CI seems random can you retry it?

PR #319, opened 2023-12-17
Cleanup menu dividers

karniv00l:
I think one is enough


mck1117:
haha how did this happen

PR #320, opened 2023-12-18
ETB calibration compatibility notes from 2022

rusefillc:


rusefillc:
I actually had that t-shirt!

image

PR #333, opened 2024-01-07
keep it simple

rusefillc:


mck1117:
You actually do want the `do...while(false)`, because that forces it to be a single statement that requires a semicolon, instead of two statements.

see https://stackoverflow.com/questions/923822/whats-the-use-of-do-while0-when-we-define-a-macro

PR #336, opened 2024-01-07
adding assertListIsSorted coverage to simulator

rusefillc:


mck1117:
Can we quantify the performance impact of this change? Those were guarded out for a reason, and that reason is that this is an extremely hot execution path. Given that we've never ever seen one of those assertions fail, I'm not sure this is worth the performance cost of adding the checks.

rusefillc:
> Can we quantify the performance impact of this change? 

Why would anyone care for _simulator_ performance impact? Whatever is simulator performance impact it's justified by additional coverage of more non-deterministic simulator.

mck1117:
> simulator

ah, missed that one word.

And it seems the compiler in the firmware case does completely remove the empty calls, so zero hit there.

PR #344, opened 2024-01-10
encapsulation

rusefillc:


rusefillc:
@mck1117 what do you think?

PR #357, opened 2024-01-31
no one uses .dfu and .hex from bundle

rusefillc:


mck1117:
cherry picked part of this, will keep hex for now

PR #360, opened 2024-02-13
put compilers in a submodule

mck1117:
Don't require compiler put on $PATH. You can now build FOME on a GitHub codespace, or in some other container with some basic build tools (`make`, etc) and that's it!

Auto detects which compiler to use for 64 bit Linux and Intel/ARM macOS.

This also makes compiler upgrades easier, as CI runners can change compiler version with the code checkout, not with any dependencies installed on the machine.


nmschulte (approved):
only suggestion is to perhaps make `TRGT` overridable

PR #361, opened 2024-02-13
High rpm above 10k

mi-hol:


firmware/config/engines/benelli.cpp
```diff
@@ -0,0 +1,137 @@
+/*
+ * @file benelli.cpp
+ *
+ * set engine_type 33
+ *
+ * @date Feb 13, 2024
+ * @author Michael Holzer, (c) 2024
+*/
+
+#include "pch.h"
```
mck1117, line 10: I would like to avoid pre-cooked presets for rare/unusual engines, as they bloat the firmware image size and have potential for tech debt headaches later. Is it possible to base your test off of a pre-existing configuration?

mi-hol: I fully understand the concern but there seems no pre-existing engine for 3-cylinder high rev motor cycles.
Yamaha, Triumph, MV Agusta and Benelli share technical details in  their motor cycle engine designs therefore this could be the foundation for a new class of engines.
I thought only this approach would assure proper testing for unusual engines early in the dev cycle.

Happy to use a different approach but would need hints/pointers

mck1117: As far as the ECU is concerned, there isn't anything special about these engines, or even odd-cylinder-count engines in general. It mostly treats any engine as N 1-cylinder engines, instead of 1 N-cylinder engine.

mi-hol: I hear what you say, just wanted to make sure a high-rev engine gets tested with a unit test.
I didn't find an existing unit test going up to 12k RPM, therefore I tried to close this gap using a new engine type. I'm more than happy to implement the requirement using a different approach.

Please advise what approach would be a better solution from your point of view?

mi-hol: https://github.com/rusefi/rusefi/commit/d3461e39051b070adeeb451d64a036a062cbc01f sounds like it could be related to my requirement but I lack the technical understanding to be sure

mck1117: That was fixed long ago and hasn't been a problem in FOME for a long time.

mck1117:
any progress to make here or should we close?

mi-hol:
> any progress to make here

well no progress was made due to above raised but unanswered question:

> I didn't find an existing unit test going up to 12k RPM, therefore I tried to close this gap using a new engine type. I'm more than happy to implement the requirement using a different approach.
> 
> Please advise what approach would be a better solution from your point of view?

Are you saying that my reasoning convinced you to consider my initial approach as valid?

mi-hol:
> well no progress was made due to above raised but unanswered question:

still no answer to my question but a bug in RE around exactly this use case  https://github.com/rusefi/rusefi/issues/6292

mck1117:
> still no answer to my question but a bug in RE around exactly this use case [rusefi/rusefi#6292](https://github.com/rusefi/rusefi/issues/6292)

that is an extreme edge case that I'm not worried about for any real scenario

PR #370, opened 2024-02-22
Add cranking taper duration multiplier based on CLT

### LINKED ISSUE #369: Add cranking IAC taper duration multiplier
I was wondering, would it be useful to have an `After cranking IAC taper duration multiplier override table`?

I enabled the `Override cranking IAC CLT multiplier` to change the IAC position based on the temperature, which really helps when the engine gets warm and doesn't get too much air when cranking.

But the `After cranking IAC taper duration` is always a fixed (in my case, 200 cycles) number. When an engine is warm, the IAC could taper down sooner because we don't need to hold the RPMs as high and for that long.

Krakert:
Fixes #369

### DIFF
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -79,8 +79,13 @@ IIdleController::Phase IdleController::determinePhase(int rpm, int targetRpm, Se
 	return Phase::Idling;
 }
 
-float IdleController::getCrankingTaperFraction() const {
-	return (float)engine->rpmCalculator.getRevolutionCounterSinceStart() / engineConfiguration->afterCrankingIACtaperDuration;
+float IdleController::getCrankingTaperFraction(float clt) const {
+    if (engineConfiguration->overrideCrankingTaperDurationSetting) {
+        return (float) engine->rpmCalculator.getRevolutionCounterSinceStart() /
+               (engineConfiguration->afterCrankingIACtaperDuration *
+                interpolate2d(clt, config->cltCrankingTaperCorrBins, config->cltCrankingTaperCorr));
+    }
+    return (float)engine->rpmCalculator.getRevolutionCounterSinceStart() / engineConfiguration->afterCrankingIACtaperDuration;
 }
 
 float IdleController::getCrankingOpenLoop(float clt) const {
@@ -309,7 +314,7 @@ float IdleController::getIdlePosition(float rpm) {
 		m_lastTargetRpm = targetRpm;
 
 		// Determine cranking taper
-		float crankingTaper = getCrankingTaperFraction();
+		float crankingTaper = getCrankingTaperFraction(clt);
 
 		// Determine what operation phase we're in - idling or not
 		float vehicleSpeed = Sensor::getOrZero(SensorType::VehicleSpeed);
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -80,12 +80,13 @@ IIdleController::Phase IdleController::determinePhase(int rpm, int targetRpm, Se
 }
 
 float IdleController::getCrankingTaperFraction(float clt) const {
-    if (engineConfiguration->overrideCrankingTaperDurationSetting) {
-        return (float) engine->rpmCalculator.getRevolutionCounterSinceStart() /
-               (engineConfiguration->afterCrankingIACtaperDuration *
-                interpolate2d(clt, config->cltCrankingTaperCorrBins, config->cltCrankingTaperCorr));
+    float taperDuration = engineConfiguration->afterCrankingIACtaperDuration;
+
+    if (engineConfiguration->useCrankingIdleTaperTableSetting) {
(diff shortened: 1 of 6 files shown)


firmware/tunerstudio/rusefi.input
```diff
@@ -3823,6 +3835,7 @@ cmd_set_engine_type_default					= "@@TS_IO_TEST_COMMAND_char@@@@ts_command_e_TS_
 		field = "Cranking base IAC position",			crankingIACposition
 		field = "After cranking IAC taper duration",afterCrankingIACtaperDuration
```
mck1117, line 3833: add a condition here to disable this setting when the new switch is turned on

firmware/integration/rusefi_config.txt
```diff
@@ -981,6 +981,7 @@ bit skippedWheelOnCam,"On camshaft","On crankshaft";Where is your primary skippe
 	bit complexWallModel,"Advanced (tables)","Basic (constants)";Should we use tables to vary tau/beta based on CLT/MAP, or just with fixed values?
 	bit alwaysInstantRpm
 	bit isMapAveragingEnabled
+	bit overrideCrankingTaperDurationSetting; If enabled, use separate temperature multiplier table for cranking taper duration.
```
mck1117, line 984: I'm not sure `override` is the right language here - maybe `useCrankingIdleTaperTable` or something would be clearer?

mck1117:
I'd love to see a unit test for the new behavior :)

firmware/tunerstudio/rusefi.input
Krakert: We still use the `afterCrankingIACtaperDuration` or do you mean something else?

firmware/integration/rusefi_config.txt
Krakert: Will change, just looked add `overrideCrankingIacSetting`

firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -79,8 +79,13 @@ IIdleController::Phase IdleController::determinePhase(int rpm, int targetRpm, Se
...
 
-float IdleController::getCrankingTaperFraction() const {
-	return (float)engine->rpmCalculator.getRevolutionCounterSinceStart() / engineConfiguration->afterCrankingIACtaperDuration;
+float IdleController::getCrankingTaperFraction(float clt) const {
+    if (engineConfiguration->overrideCrankingTaperDurationSetting) {
+        return (float) engine->rpmCalculator.getRevolutionCounterSinceStart() /
+               (engineConfiguration->afterCrankingIACtaperDuration *
+                interpolate2d(clt, config->cltCrankingTaperCorrBins, config->cltCrankingTaperCorr));
+    }
+    return (float)engine->rpmCalculator.getRevolutionCounterSinceStart() / engineConfiguration->afterCrankingIACtaperDuration;
```
mck1117, line 88: I would avoid duplicating this code if you can - it's not obvious to the reader why these two cases are different when they look so similar on first glance.

prefer something like

```c++
float x = config.taperDuration;

if (new_behavior) {
    x *= table_lookup();
}

return getRevCounter() / x;
```

mck1117: that style of writing it makes it very clear which parts are the same, and which parts are different when the setting is enabled

firmware/integration/rusefi_config.txt
```diff
@@ -1556,6 +1557,9 @@ uint8_t[PEDAL_TO_TPS_SIZE] autoscale pedalToTpsRpmBins;;"RPM", 100, 0, 0, 25000,
 float[CLT_CRANKING_CURVE_SIZE] cltCrankingCorrBins;CLT-based cranking position multiplier for simple manual idle controller;"C", 1, 0, -100, 250, 2
 float[CLT_CRANKING_CURVE_SIZE] cltCrankingCorr    ;CLT-based cranking position multiplier for simple manual idle controller;"%", 1, 0, 0, 500, 2
 
+float[CLT_CRANKING_CURVE_SIZE] cltCrankingTaperCorrBins;CLT-based taper duration multiplier for simple manual idle controller;"C", 1, 0, -100, 250, 2
+float[CLT_CRANKING_CURVE_SIZE] cltCrankingTaperCorr    ;CLT-based taper duration multiplier for simple manual idle controller;"%", 1, 0, 0, 500, 2
```
mck1117, line 1561: 1. is 500 a reasonable maximum value for this parameter? maybe 5 is safer?
2. using float here is a waste of storage, maybe int8_t for temp, and uint8_t with scale 0.02 for multiplier would be better?

Krakert: Maybe a good idea to also make these change this for the `cltCrankingCorr`

mck1117: yes, finding other tables we can compact would be a great follow-on task #371

Krakert:
> I'd love to see a unit test for the new behavior :)

Added a unit test

mck1117 (approved):

firmware/tunerstudio/rusefi.input
mck1117: oop, yeah, it's a multiplier on that duration, not just the duration. lgtm.

PR #373, opened 2024-03-01
implement TS console over ATWINC1500 WiFi

mck1117:
Implement WiFi driver and enable for Atlas


rusefillc:
Nice! Does this HW exist? What data rate are you getting?

mck1117:
> Nice! Does this HW exist? 

Nope, not other than a nucleo on the bench with jumper wires.

> What data rate are you getting?

There are some further optimizations to make, but on an F4 (not running at the full supported SPI speed) I was getting 150-175hz for full updates in TS of ~1500 bytes or whatever it is.

rusefillc:
@mck1117 very cool stuff!

PR #377, opened 2024-03-09
KnockController: fix knock retard logic

### LINKED ISSUE #246: Knock retard not activating (20230516 release)
Based on the testing we did at silverstone it seems the knock retard is not activating, knock count increases, and the volume on the sensors passes the threshold but no timing is removed. 

Over the day we tried several combinations of aggression (5 and 10%), threshold (lowered it by 2dBv) and max allowed retard (tried several tables then flat mapped) with no change in result. 

Logs and tune attached.  
MLV knock polynomial is " (-0.000000814*[RPM]*[RPM]) + (0.00986*[RPM]) - 37 + 2 "

[re_17.zip](https://github.com/FOME-Tech/fome-fw/files/12640948/re_17.zip)

nmschulte:
fixes #246

----

use method to determine maximum retard, versus the member which is unused (in favor of max knock retard table)

### DIFF
--- a/firmware/controllers/engine_cycle/knock_controller.cpp
+++ b/firmware/controllers/engine_cycle/knock_controller.cpp
@@ -74,7 +74,7 @@ bool KnockControllerBase::onKnockSenseCompleted(uint8_t cylinderNumber, float db
 			// Adjust knock retard under lock
 			chibios_rt::CriticalSectionLocker csl;
 			auto newRetard = m_knockRetard + retardAmount;
-			m_knockRetard = clampF(0, newRetard, m_maximumRetard);
+			m_knockRetard = clampF(0, newRetard, getMaximumRetard());
 		}
 	}
 


mck1117:
the member is set in `onFastCallback()`, no?

mck1117:
Doing it in `onFastCallback()` also avoids doing a potentially expensive table lookup in the hot path that gets called for every ignition event.

rusefillc:
@nmschulte any interest to TDD? maybe test proving issue as first PR and flipping text while fixing in another PR?

nmschulte:
> @nmschulte any interest to TDD? maybe test proving issue as first PR and flipping text while fixing in another PR?

That's a good idea; I need to learn more about ChibiOS and the firmware architecture at large before I can go much further.  Will close this PR as it's not the right solution.

rusefillc:
> That's a good idea; I need to learn more about ChibiOS and the firmware architecture at large before I can go much further.

Zero ChibiOS at https://github.com/FOME-Tech/fome-fw/tree/master/unit_tests yet extremely useful!

PR #391, opened 2024-03-18
better file name to actually reflect what the content is about

rusefillc:
Sibling of https://github.com/rusefi/rusefi/pull/6226


mck1117:
I think I would prefer something like `tunerstudio.template.ini` to reflect that it's still mostly an ini file. That would also allow us to remove the file assocation from `.vscode/settings.json` to pretend `.input` files are `.ini` files.

rusefillc:
> I think I would prefer something like `tunerstudio.template.ini` 

Done

rusefillc:
We did good!

PR #399, opened 2024-03-22
-Werror=stringop-overflow

rusefillc:
`if you get ... fault at a particular optimization level, you're probably relying on undefined behavior somewhere`


nmschulte:
This seems to be a false positive issued by GCC; some of this is documented here: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wstringop-overflow

Simply re-arranging the private members of `static_vector` so that `m_size` is after `m_storage` allows the Simulator build to pass with the warning treated as error.

PR #406, opened 2024-04-09
angle domain is more natural for scheduler validation

rusefillc:


mck1117:
Looks like a dup now?

PR #409, opened 2024-04-09
Lua idle rpm adder

dynfer:


firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -34,6 +34,7 @@ int IdleController::getTargetRpm(float clt) {
 
 	auto target = targetRpmByClt + targetRpmAcBump;
 	idleTarget = target;
+	target += luaAddRpm;
```
mck1117, line 37: should these two lines be swapped, so the idle logging shows the target including lua bump?

dynfer: valid point currently it wont show the lua bump anywhere

rusefillc:
@dynfer wow you've reinvented some code I wrote, what a coincidence!

PR #415, opened 2024-04-28
use a custom chrono-like type for time

mck1117:


nmschulte:
Whut's it resolve/improve?  Will look closer tomorrow I hope.  :+1:

rusefillc:
@mck1117 why not real chrono like at https://github.com/rusefi/libfirmware/pull/16?

mck1117:
> @mck1117 why not real chrono like at [rusefi/libfirmware#16](https://github.com/rusefi/libfirmware/pull/16)?

This is a better stepping stone as we can control implicit operators to ease the transition.

Issue #417, opened 2024-04-30
m_maximumretard broken, resulting in knock controller not working. Related to Closed Loop Fueling?


rk-iv:
Proteus f7 BMM ecu for Mazda Miata NA6

When setting up the Max knock retard table with an array of values ranging from 2 through 10. The variable m_maximumretard does not correlate properly, I assumed if i set up a table of RPM, MAP, Knock: max retard i would see a table that represents my tune table. That was not the case. Furthermore, It made me realize it was not pulling timing when knock count incremented. 

This is what I have found so far

- ECU powered on W/ Closed Loop Fuel Control - ON
              - m_maximumretard is a constant 0 regardless of Max Knock Retard Table
- ECU powered on W/ Closed Loop Fuel Control - ON, then Turned OFF
              -   m_maximumretard changes from 0 to 2 regardless of Max Knock Retard Table
              - Upon power cycle, m_maximumretard value stayed 2
- ECU powered on w/ Closed Loop Fuel Control ON, then OFF, Then ON again
              - m_maximumretard starts at 0, goes to 2.
              - Started the car, m_maximumretard goes to 8, stays at 8 constant regardless of Max Knock Retard Table 
              - Drove around, took a log, both Closed loop fuel control and Knock Controller work as Expected. Pulling timing, and fuel trimming.
              - When I turned the car off, the m_maximumretard stayed constant at 8. When I power cycled ECU, value returned to 0.

nmschulte:
see also https://github.com/FOME-Tech/fome-fw/issues/246

rk-iv:
M_maximumretard constant zero when ecu is powered on with Closed loop fuel control

![Screenshot 2024-04-30 094508](https://github.com/FOME-Tech/fome-fw/assets/165811927/6bf198bc-612c-4ac1-b599-f6af12e6a6af)

m_maximumretard changes to to constant 2 when CLFC is turned Off.

![Screenshot 2024-04-30 094539](https://github.com/FOME-Tech/fome-fw/assets/165811927/e9951fb4-6e6a-402d-aa02-6ff8ca585dab)

Turned CLFC back on, and turned on car, Knock controller is Active

![Screenshot 2024-04-30 094620](https://github.com/FOME-Tech/fome-fw/assets/165811927/8f636912-f10d-418c-91f8-5bbb59579c81)


[...]

PR #419, opened 2024-05-02
fix knock retard table init

### LINKED ISSUE #246: Knock retard not activating (20230516 release)

nmschulte:
column and row definition were swapped w/re: value lookup

fixes #246, #417

----

Andrey was right: https://github.com/FOME-Tech/fome-fw/issues/417#issuecomment-2087816779
> is this https://github.com/rusefi/rusefi/issues/6370?

Comparison to boost; RPM is columns:
2024-05-02_02h00m55s_screenshot

boost:
```cxx
// init
boostMapOpen.init(config->boostTableOpenLoop, config->boostTpsBins, config->boostRpmBins);
boostMapClosed.init(config->boostTableClosedLoop, config->boostTpsBins, config->boostRpmBins);

// getValue
float openLoop = luaOpenLoopAdd + m_openLoopMap->getValue(rpm, driverIntent.Value);
float target = m_closedLoopTargetMap->getValue(rpm, driverIntent.Value);
```

knock:
```cxx
// init
m_maxRetardTable.init(config->maxKnockRetardTable, config->maxKnockRetardRpmBins, config->maxKnockRetardLoadBins);

// getValue
return m_maxRetardTable.getValue(Sensor::getOrZero(SensorType::Rpm), getIgnitionLoad());
```

### DIFF
--- a/firmware/controllers/engine_cycle/knock_controller.cpp
+++ b/firmware/controllers/engine_cycle/knock_controller.cpp
@@ -8,14 +8,14 @@
 #include "pch.h"
 #include "knock_logic.h"
 
-void KnockController::init(engine_configuration_s const * const previousConfig) {
+void KnockController::init() {
 	m_maxRetardTable.init(config->maxKnockRetardTable, config->maxKnockRetardLoadBins, config->maxKnockRetardRpmBins);
 }
 
 void KnockController::onConfigurationChange(engine_configuration_s const * previousConfig) {
 	KnockControllerBase::onConfigurationChange(previousConfig);
 
-	init(previousConfig);
+	init();
 }
 
 int getCylinderKnockBank(uint8_t cylinderNumber) {
@@ -167,5 +167,5 @@ void Engine::onSparkFireKnockSense(uint8_t cylinderNumber, efitick_t nowNt) {
 }
 
 void initKnockCtrl() {
-	engine->module<KnockController>().unmock().init(engineConfiguration);
+	engine->module<KnockController>().unmock().init();
 }
--- a/firmware/controllers/engine_cycle/knock_controller.cpp
+++ b/firmware/controllers/engine_cycle/knock_controller.cpp
@@ -8,10 +8,14 @@
 #include "pch.h"
 #include "knock_logic.h"
 
+void KnockController::init(engine_configuration_s const * const previousConfig) {
+	m_maxRetardTable.init(config->maxKnockRetardTable, config->maxKnockRetardLoadBins, config->maxKnockRetardRpmBins);
+}
+
 void KnockController::onConfigurationChange(engine_configuration_s const * previousConfig) {
 	KnockControllerBase::onConfigurationChange(previousConfig);
 
-	m_maxRetardTable.init(config->maxKnockRetardTable, config->maxKnockRetardLoadBins, config->maxKnockRetardRpmBins);
+	init(previousConfig);
 }
 
 int getCylinderKnockBank(uint8_t cylinderNumber) {
@@ -160,4 +164,8 @@ void Engine::onSparkFireKnockSense(uint8_t cylinderNumber, efitick_t nowNt) {
 #else
 	UNUSED(nowNt);
 #endif
+}
+
+void initKnockCtrl() {
+	engine->module<KnockController>().unmock().init(engineConfiguration);
 }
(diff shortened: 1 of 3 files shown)


rusefillc:
While you are fixing this defect root cause is order of arguments on the method. See rusefi changes flipping it.

nmschulte:
> fixes #246, #417

I haven't confirmed this; this statement might be an overreach.

I'm also not sure this mistake explains how `m_maximumRetard` would be zero if the table is entirely non-zero.

----

> While you are fixing this defect root cause is order of arguments on the method. See rusefi changes flipping it.

I'll check that out soon; had looked at it just a little before.

As well, the type system should be able to save us from mistakes like this somehow.  "traits" naively come to mind but I'm admittedly unfamiliar; the standard kit (pre-traits) should be able to do it though too if I'm not mistaken.

nmschulte:
@rk-iv has effected this change by altering the TS `.ini`, like so; remapping the bins and the table axes so the `m_maxRetardTable->init` logic does what it should and TS editor can be used as intended w/o brain games:

```diff
 maxKnockRetardTable = array, U08, 19556, [6x6], "deg", 0.25, 0, 0, 30, 2
-maxKnockRetardLoadBins = array, U08, 19592, [6], "%", 1, 0, 0, 250, 0
-maxKnockRetardRpmBins = array, U08, 19598, [6], "RPM", 100.0, 0, 0, 25000, 0
+maxKnockRetardRpmBins = array, U08, 19592, [6], "%", 1, 0, 0, 250, 0
+maxKnockRetardLoadBins = array, U08, 19598, [6], "RPM", 100.0, 0, 0, 25000, 0
```

```diff
     table = maxKnockRetardTbl, maxKnockRetardMap, "Max knock retard",    1
-        xBins        = maxKnockRetardRpmBins, RPMValue
-        yBins        = maxKnockRetardLoadBins, ignitionLoad
+        xBins        = maxKnockRetardLoadBins, RPMValue
+        yBins        = maxKnockRetardRpmBins, ignitionLoad
         zBins        = maxKnockRetardTable
         gridOrient  = 250,    0, 340 ; Space 123 rotation of grid in degrees.
```

It doesn't fully solve the problem; the CLFC off->on dance is still required to achieve desired effect.

nmschulte:
I found that `KnockController` module had no "initialize" method, and so `m_maximumRetard` was not being initialized on ECU reset.  Added an implementation similar to that of `BoostController`; probably not perfect but should resolve the issue.

firmware/controllers/engine_cycle/knock_controller.cpp
```diff
@@ -8,10 +8,14 @@
 #include "pch.h"
 #include "knock_logic.h"
 
+void KnockController::init(engine_configuration_s const * const previousConfig) {
+	m_maxRetardTable.init(config->maxKnockRetardTable, config->maxKnockRetardLoadBins, config->maxKnockRetardRpmBins);
```
nmschulte, line 12: note this flips the load/rpm bins to match the current implementation

rusefillc:
> I found that KnockController module had no "initialize" method, and so m_maximumRetard was not being initialized on ECU reset.

if that's the case, and it seems to be the case, I assume the most important next step is to make sure such a defect never occurs again? https://github.com/rusefi/rusefi/issues/6461

firmware/controllers/engine_cycle/knock_controller.cpp
```diff
@@ -8,10 +8,14 @@
 #include "pch.h"
 #include "knock_logic.h"
 
+void KnockController::init(engine_configuration_s const * const previousConfig) {
```
rusefillc, line 11: what is the point of 'previousConfig' here?

nmschulte: you are correct, I was moving too fast, didn't realize `config` was "global" and `previousConfig` is not what `onConfigurationChange` was/is using

nmschulte:
> I assume the most important next step is to make sure such a defect never occurs again?

I agree, see also #422 (and #421).

firmware/controllers/engine_cycle/knock_controller.cpp
mck1117: not the minimal diff to fix the bug, sigh

PR #420, opened 2024-05-02
fix building firmware on linux

nmschulte:
`--processor` is non-portable, and systems like Kali and Debian do not play games like Ubuntu et al which e.g. patch to `--machine`

~~Also support Linux AArch64 (arm64) host~~

### DIFF
--- a/firmware/use_arm_gcc.mk
+++ b/firmware/use_arm_gcc.mk
@@ -1,12 +1,15 @@
 ifeq ($(TRGT),)
-	UNAME_S := $(shell uname -s)
 	UNAME_SM := $(shell uname -sm)
 $(info UNAME_SM:          $(UNAME_SM))
+	KERNEL := $(firstword $(UNAME_SM))
+	MACHINE := $(lastword $(UNAME_SM))
 
-	ifeq ($(UNAME_S),Darwin)
+	ifeq ($(KERNEL),Darwin)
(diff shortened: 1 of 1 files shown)


firmware/use_arm_gcc.mk
```diff
@@ -1,30 +1,35 @@
...
 
-	ifeq ($(UNAME_S),Darwin)
+	ifeq ($(KERNEL),Darwin)
 		COMPILER_PLATFORM = arm-gnu-toolchain-11.3.rel1-darwin-x86_64-arm-none-eabi
-	else ifeq ($(UNAME_SP),Linux x86_64)
-		COMPILER_PLATFORM = arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi
+	else ifeq ($(KERNEL),Linux)
+		ifeq ($(MACHINE),x86_64)
+			COMPILER_PLATFORM = arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi
+		else ifeq ($(MACHINE),aarch64)
```
mck1117, line 12: These files don't actually exist in the toolchain dirs, so let's not support it here yet

nmschulte:
Ah, my bad, I thought it downloaded on the fly from Arm 👍 -- will remove that commit soon.
-------- Original Message --------

PR #426, opened 2024-05-15
idle stepper TS panel disable should not depend ETB functions

nmschulte:
it seems https://github.com/rusefi/rusefi/pull/4904/files may have been merged without thorough review

does this change make sense?  configuring idle stepper hardware should not depend on ETB function/configuration, right?

also, perhaps still an issue exists if both ETB and idle stepper are trying to use the same hardware?


nmschulte:
rusefi/rusefi#5972 seems to be this issue; also: https://rusefi.com/forum/viewtopic.php?p=49196#p49196

PR #427, opened 2024-05-23
Mini (R52) CAN dash

nmschulte:


firmware/controllers/can/can_dash.cpp
```diff
@@ -1316,4 +1321,49 @@ void canDashboardAim(CanCycle cycle) {
...
+    // 0x316 -- needs at least every 500ms
+    // 0x61a -- needs at least every 1000ms?
+    // 0x61f -- needs at least every 1000ms?
+
+    int rpm = Sensor::getOrZero(SensorType::Rpm);
+
+    if (cycle.isInterval(CI::_50ms)) {
+        {   // block scope to invoke CanTxMessage deconstructor (send)
+            CanTxMessage msg(CAN_MINI_DME1_ID, 8);
+            msg.setShortValue(rpm * 6.4, 3);
```
nmschulte, line 1334: In my tests, this was indeed big-endian vs little-endian as others have noted.  I need to confirm with a second unit and some closer analysis.

firmware/controllers/can/can_dash.cpp
```diff
@@ -1316,4 +1321,49 @@ void canDashboardAim(CanCycle cycle) {
...
+                    msg.setBit(4, 6);
+                }
+
+                if (rpm >= 7000) {
+                    // enable red "7-8k" LED bars
+                    msg.setBit(4, 7);
+                }
+            }
+        }
+    } else if (cycle.isInterval(CI::_500ms)) {
```
nmschulte, line 1360: this will never hit ... (50 divides 500 evenly ...)

PR #429, opened 2024-05-23
stringop-overflow error

### LINKED ISSUE #400: stringop-overflow is a valuable GCC error
stringop-overflow seems to be the most serious ignored error https://github.com/FOME-Tech/fome-fw/blob/master/firmware/rusefi_rules.mk 

Would be better not to ignore it but at the moment simulator fails to compile

See also https://github.com/FOME-Tech/fome-fw/pull/399/files https://github.com/rusefi/rusefi/issues/6270

nmschulte:
resolves #400 (see also #399)


rusefillc:
Nice!

PR #430, opened 2024-05-24
Toyota ETCS-i [01/02]: meld FordRedundantTps Sensor into RedundantSensor

nmschulte:
Meld `FordRedundantTps` `Sensor` into `RedundantSensor`:
- `RedundantSensor` accepts `secondMaximum`
  - TPS and PPS default to fully redundant (`100`)
- improve `Sensor` info printing (`showInfo`)
- cleanup `de`/`initTps`
  - remove `isFord` orthnormality
- meld unit tests


nmschulte:
I have not tested these changes on actual hardware since the initial implementation against rusEFI; I'm not sure when I'll have time to do this, but I see no reason currently it should work any different.

nmschulte:
> I have not tested these changes on actual hardware since the initial implementation against rusEFI; I'm not sure when I'll have time to do this, but I see no reason currently it should work any different.

I have tested this with an ECTS-i ETB; it works well.

firmware/controllers/sensors/redundant_sensor.cpp
```diff
@@ -9,33 +9,66 @@ RedundantSensor::RedundantSensor(SensorType outputType, SensorType first, Sensor
...
+	m_secondMaximum = secondMaximum;
 }
 
 SensorResult RedundantSensor::get() const {
-	auto result1 = Sensor::get(m_first);
+	// Sensor 1 is always full range, i.e. 0% -> 100%
+	auto sensor1 = Sensor::get(m_first);
+
+	// If sensor 1 result is invalid, return inconsistency error
+	if (!sensor1.Valid) {
```
mck1117, line 23: I don't think it's correct to put this above the `m_ignoreSecond` check - in single sensor mode, we want to return the true error instead of `Inconsistent`

nmschulte: I updated the unit tests to expect this; they should fail here momentarily and subsequent commit will fix this behavior.  Good catch; thank you.

nmschulte: https://github.com/FOME-Tech/fome-fw/actions/runs/9354363787/job/25747069591?pr=430#step:8:25135
```
[==========] 595 tests from 120 test suites ran. (387 ms total)
[  PASSED  ] 592 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] SensorRedundantIgnoreSecond.OnlyFirst
[  FAILED  ] SensorRedundantIgnoreSecond.OnlySecond
[  FAILED  ] SensorRedundantIgnoreSecond.SetBothIgnoreSecond
```

PR #432, opened 2024-05-24
simulator: more arguments

nmschulte:
- use [Argp](https://www.gnu.org/software/libc/manual/html_node/Argp.html) to parse arguments
- allow to specify SocketCAN device (`--socketcan-device`/`-d`)
- make timeout argument an option (`--timeout`/`-t`)
- allow to disable verbosity (`--quiet`/`-q`)

### DIFF
--- a/simulator/libc_argp.h
+++ b/simulator/libc_argp.h
@@ -6,22 +6,23 @@ static char doc[] =
     "FOME Simulator -- https://wiki.fome.tech/";
 
 /* A description of the arguments we accept. */
-static char args_doc[] = "[TIMEOUT]";
+static char args_doc[] = "";
 
 /* The options we understand. */
 static struct argp_option options[] = {
     {"quiet", 'q', 0,         0, "Don't produce verbose output", 0 },
     {"socketcan-device",
                 'd', "DEVICE",  0, "SocketCAN DEVICE (default: can0) to use", 0 },
+    {"timeout", 't', "SECONDS", 0, "Run for SECONDS and then exit (negative values ignored)", 0 },
     { 0, 0, 0, 0, 0, 0 }
 };
 
 /* Used by main to communicate with parse_opt. */
 struct arguments
 {
-    int timeout;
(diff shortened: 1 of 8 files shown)


mck1117:
>     * retain current "timeout" argument behavior

I wouldn't mind changing that behavior (but also change ci) to use a flag instead of a bare arg

nmschulte:
> I wouldn't mind changing that behavior (but also change ci) to use a flag instead of a bare arg

done

simulator/libc_argp.h
```diff
@@ -0,0 +1,65 @@
...
+            break;
+        case 'q':
+            arguments->quiet = 1;
+            break;
+        case 't':
+            arguments->timeout = atoi(arg);
+            break;
+
+        case ARGP_KEY_ARG:
+            /* if (state->arg_num >= 0) */ {
```
nmschulte, line 48: wanted to retain this format in case we add arguments later, but the comparison results in compile error due to `unsigned int` comparison w/ `0` is always true.

```
libc_argp.h:48:32: error: comparison of unsigned expression in ‘>= 0’ is always true [-Werror=type-limits]
   48 |             if (state->arg_num >= 0) {
```

simulator/simulator/rusEfiFunctionalTest.cpp
```diff
@@ -84,7 +84,7 @@ static void runChprintfTest() {
 
 }
 
-void rusEfiFunctionalTest(void) {
+void rusEfiFunctionalTest(char const * const socketcanDevice) {
 	printToConsole("Running rusEfi simulator version:");
```
nmschulte, line 88: wish there was a `printfToConsole` ;)

mck1117: you mean to the FOME console? that exists, it's called `efiPrintf`

mck1117:
uhh, looks like this doesn't work on the windows build...

PR #437, opened 2024-06-01
add knock retard aggression value advise

nmschulte:


mck1117:
I think you meant `advice` with a c

PR #442, opened 2024-06-14
remove functionpointer sensor

nmschulte:
This seems unused and referring to a legacy pattern.  Obviously decline if we want to keep it around for w/e reason.


mck1117:
lgtm

PR #452, opened 2024-07-23
Engine modules participate in the build system

mck1117:
Add an engine module simply by including its makefile. This makes it easier to compose an ECU out of a set of modules, with less dependence on complicated preprocessor definitions.


firmware/console/status_loop.cpp
```diff
@@ -424,13 +427,15 @@ static void updateFuelCorrections() {
 }
 
 static void updateFuelResults() {
+	#ifdef MODULE_TRIP_ODO
```
nmschulte, line 430: nit: usually we're always left-aligning macros

nmschulte, line 430: nit: `ifdef` vs `if` -- there are ~10x more `if` vs `ifdef` already; do we care?  my brain just keeps saying `clang-format`  :/

mck1117: they do different things - ifdef happens if the macro merely exists, and if happens if the thing is truthy

mck1117: it's simpler to define the symbol when the module is included, instead of having to define all not-loaded modules to 0, then re-define some to 1 if loaded

Issue #454, opened 2024-07-25
custom filters for autotune


BeerMoneyMotorsports:
Add custom filters option for autotune like MS has.  

![image](https://github.com/user-attachments/assets/1610a526-d319-4e26-a0d5-c8f6445609fe)
![image](https://github.com/user-attachments/assets/c4d53115-9ad7-4535-981e-4cc7d8423b76)
![image](https://github.com/user-attachments/assets/e190fb70-b0fd-4297-92d0-a2813e631019)

mck1117:
I'm not sure how to do that, can you find an ini that supports it so I can analyze?

nmschulte:
Interestingly, the Warmup Enrichment tuning shows this Custom Filter button with FOME.
![image](https://github.com/user-attachments/assets/e4e42064-a7de-4087-a860-e0bad961273b)

PR #455, opened 2024-07-25
enable VE Analyze custom filter

nmschulte:
for #454


nmschulte:
http://www.efianalytics.com/MegaLogViewer/formulas.html
(http://www.tunerstudio.com/index.php/support/manuals/88-math-parser-functions)

PR #462, opened 2024-08-13
execute critical tasks on one big thread

mck1117:
Move some separate functions to all be called from a single fast loop. The loop runs at 1khz, then executes various things at integer ratios off of that speed. Right now that means ADC at 500hz, ETB at 500hz, periodic fast at 250hz, periodic slow at 20hz.

- Saves thread stack size for things running at similar priority at similar rates
- Avoids any phase lag or phase noise between ADC measurements and things using it (ADC is sampled, then updates are immediately run)
- Static checking to ensure rates in use are valid for the loop rate


rusefillc:
Nice!

PR #463, opened 2024-08-15
Update rusefi_config.txt

DenisMitchell:
Corrected and added more detail to the tooltip for the ETB Bias Curve value.


mck1117:
LGTM!

Issue #358, opened 2024-02-02
Transmission Gear Ratio presets


BeerMoneyMotorsports:
Implement a dropdown box for some common transmissions.  

Miata NA/NB 5 speed
Miata NB 6 sped
Miata NC 5 speed
Miata NC 6 speed
350z 6 speed
BMW Getrag e30 5 speed
BMW E36/46 ZF5 5 speed
BMW E36/E46 Getrag 5 Speed
BMW e46 ZF6 6 speed
etc

https://kpower.industries/pages/gearing-calculator#%5B225%2C45%2C15%2C8000%2C%223.909%22%2C%223.000%22%2C70%2C%22Miata%206-speed%22%2C3.76%2C2.269%2C1.646%2C1.257%2C1%2C0.843%5D

G0nz0w:
Can I request dropdown addition for Lexus IS200 J160 gear box?

Final drive ratio - 3.90 (Manual Limited Slip Diff)
                          - 4.1 (Automatic Diff)
Forward gear count - 6

1st gear ratio - 3.874
2nd gear ratio - 2.175
3rd gear ratio - 1.484
4th gear ratio - 1.223
5th gear ratio - 1.000
6th gear ratio - 0.869

PR #471, opened 2024-08-25
Added gear ratio presets for common transmissions in Tunerstudio

alrijleh:
Addresses the enhancement requested in https://github.com/FOME-Tech/fome-fw/issues/358


alrijleh:
Why did the build/checks fail for unrelated firmware?

mck1117:
> Why did the build/checks fail for unrelated firmware?

It was broken on master - I've since fixed it and this looks good.

Issue #451, opened 2024-07-20
address MAP Averaging TODOs


nmschulte:
`map_averaging.cpp` has a number of TODOS to improve functionality and performance:
- `startAveraging` (called from scheduled `mapAveragingTriggerCallback`/`scheduleByAngle`)
  - ```cpp
    // TODO: set currentMapAverager based on cylinder bank
    auto& averager = getMapAvg(currentMapAverager);
    averager.start();
    ```
- `MapAverager::stop`
  - ```cpp
    // TODO: this should be per-sensor, not one for all MAP sensors
    averagedMapRunningBuffer[averagedMapBufIdx] = averageMap;
    // increment circular running buffer index
    averagedMapBufIdx = (averagedMapBufIdx + 1) % mapMinBufferLength;
    // find min. value (only works for pressure values, not raw voltages!)
    float minPressure = averagedMapRunningBuffer[0];
    for (int i = 1; i < mapMinBufferLength; i++) {
        if (averagedMapRunningBuffer[i] < minPressure)
            minPressure = averagedMapRunningBuffer[i];
    }

    setValidValue(minPressure, getTimeNowNt());
    ```
- `refreshMapAveragingPreCalc`
  - ```cpp
    for (size_t i = 0; i < engineConfiguration->cylindersCount; i++) {
        angle_t cylinderOffset = getEngineCycle(getEngineRotationState()->getOperationMode()) * i / engineConfiguration->cylindersCount;
        efiAssertVoid(ObdCode::CUSTOM_ERR_MAP_CYL_OFFSET, !cisnan(cylinderOffset), "cylinderOffset");
        // part of this formula related to specific cylinder offset is never changing - we can
        // move the loop into start-up calculation and not have this loop as part of periodic calculation
        // todo: change the logic as described above in order to reduce periodic CPU usage?
        float cylinderStart = start + cylinderOffset - offsetAngle + tdcPosition();
        wrapAngle(cylinderStart, "cylinderStart", ObdCode::CUSTOM_ERR_6562);
        engine->engineState.mapAveragingStart[i] = cylinderStart;
    }
    engine->engineState.mapAveragingDuration = interpolate2d(rpm, c->samplingWindowBins, c->samplingWindow);
    ```
- `mapAveragingTriggerCallback`
  - ```cpp
    // todo: this could be pre-calculated
    int samplingCount = engineConfiguration->measureMapOnlyInOneCylinder ? 1 : engineConfiguration->cylindersCount;

[...]

PR #475, opened 2024-08-29
MAP avg uses engine module callbacks

mck1117:


nmschulte:
related/nearby: https://github.com/FOME-Tech/fome-fw/issues/451

PR #481, opened 2024-09-03
Update README.md - link to developer.md instead of RusEFI forum

DenisMitchell:
Link to developer.md instead of RusEFI forums; instructions on the RusEFI forums do not appear to apply to FOME at this point.

Added a direct link to Releases further up, outside of the Contributing/Developing section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,6 @@ Online and offline (PDF) versions of the user manual available here: https://wik
  * [Unit Tests](/unit_tests) - Unit tests of firmware
  * [Misc tools](/java_tools) - Misc development utilities
  * [misc/Jenkins](/misc/jenkins) - Continuous integration scripts
- * [Releases](https://github.com/FOME-Tech/fome-fw/releases) - Pre-built firmware releases
 
 # External Links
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,7 @@ Online and offline (PDF) versions of the user manual available here: https://wik
  * [Unit Tests](/unit_tests) - Unit tests of firmware
  * [Misc tools](/java_tools) - Misc development utilities
  * [misc/Jenkins](/misc/jenkins) - Continuous integration scripts
+ * [Releases](https://github.com/FOME-Tech/fome-fw/releases) - Pre-built firmware releases
```
nmschulte, line 35: one of these things is not like the other :/

DenisMitchell: Is there a way to link to releases without a full https:// link? I tried a couple things but could only find the explicit reference as was used at the bottom of the file.

nmschulte: Sorry, I should have been more clear: the new bullet is not like the others; the others tell about the content that is source-controlled in the repository -- the new one is describing something else.

I think it's good to have, but not as part of the "what's in here?" list?  I'm probably being picky, sorry.

and, in fact, there's already a link to the releases under "Release Notes" heading in the last line of the file

DenisMitchell: No worries, I'm still learning with all of this. My thought when adding that was to put the releases link a bit higher up for people who are new to FOME, but in fact it's linked right now from the home page of fome.tech so I'm thinking a second link in readme.md would be redundant. I'm imagining most people will probably end up at fome.tech before they end up on the repo.

DenisMitchell:
I'm guessing updating readme.md shouldn't have an effect on the failing check above. Looks like `set ignition_mode 1` took the full time-out period and failed to return data prior to `reset_engine_chart` timing out fully and failing.

PR #483, opened 2024-09-05
Update tachometer.cpp - fix tach sweep

DenisMitchell:
fix tachsweep - x2 and y2 were transposed in interpolateClamped lines.


mck1117:
oh good spot, that'll do it

Issue #478, opened 2024-08-31
Vary DFCO MAP threshold based on current RPM


alrijleh:
I've noticed with my car (NA Miata) that at 4000 rpm the car will decelerate at around 20kpa, while at 2000 it will be closer to 30.  In my experience, setting the DFCO threshold at 24 will result in the car never entering DFCO at low engine speeds, but 28 will result in a jolt when fuel cuts at high engine speed.

Adding an option to set the DFCO MAP threshold on a curve based on RPM would solve the issue, allowing for fuel saving at low engine speed, and smoother behavior at high engine speeds.

DenisMitchell:
Can this be avoided by using the DFCO delay?

alrijleh:
I currently have a small amount of DFCO delay (0.3 seconds) in place, and still observe this behavior.  Another different approach to improving DFCO would be adding hysteresis to the MAP threshold like we do with RPM. ie cut below 22 MAP and enable fuel above 25 MAP

PR #485, opened 2024-09-06
configure DFCO MAP threshold to vary based on RPM

alrijleh:
Addresses my idea proposed in https://github.com/FOME-Tech/fome-fw/issues/478

I did a brief test drive in my car to verify that it works.  Feel free to reject if this seems like it would contribute unnecessary bloat to the project.
Capture


mck1117:
1. looks like a merge conflict in the TS template
2. I'd love to see a unit test for this new behavior!

alrijleh:
On it.
Do you think it would make sense to also add a field to set hysterics for the MAP threshold?

alrijleh:
Also, do you think the “ MAP ModE” drop-down selector is needed?  Fixed is the same as setting a flat table, but it allows the existing tune settings to work when updating to the new firmware for the first time.

mck1117:
> Also, do you think the “ MAP Mode” drop-down selector is needed? Fixed is the same as setting a flat table, but it allows the existing tune settings to work when updating to the new firmware for the first time.

Yes - breaking back compat is sort of unacceptable, so the dropdown is required.

Issue #22, opened 2023-02-23
Option to disable DFCO if clutch pressed


mck1117:
Don’t DFCO during shifts as it can cause transient knock

lots of OEM ECU have this option

https://github.com/rusefi/rusefi/issues/4923

PR #487, opened 2024-09-16
Disable DFCO when clutch is applied

alrijleh:
Allow clutchUp switch to be a gating term for fuel cut activation.  Enables feature requested in https://github.com/FOME-Tech/fome-fw/issues/22


firmware/controllers/sensors/sensor_type.h
```diff
@@ -71,6 +71,8 @@ enum class SensorType : unsigned char {
 
 	BarometricPressure,
 
+	ClutchUp,
```
mck1117, line 74: unrelated?

firmware/controllers/algo/defaults/default_fuel.cpp
```diff
@@ -109,6 +109,7 @@ static void setDefaultFuelCutParameters() {
 	engineConfiguration->coastingFuelCutTps = 2;
 	engineConfiguration->coastingFuelCutMap = 30;
 	engineConfiguration->coastingFuelCutClt = 60;
+	engineConfiguration->disableFuelCutOnClutch = false;
```
mck1117, line 112: By the way, you don't have to set any defaults that have a bit value of 0. The configuration is zeroed before reset, so everything defaults to 0/false/etc.

firmware/controllers/sensors/sensor_type.h
alrijleh: Good catch.  Initially thought I needed to plumb it through before I saw that the digital inputs were getting set in engine.cpp.  This can/should be removed

PR #491, opened 2024-09-18
Consoldate Tunerstudio panels

alrijleh:
The dropdown submenus are getting long with the recently added features, any opposition to combining related tables and settings?

Menus affected in this PR:
- After start enrichment
- cranking+priming pulse
- accel enrichment + tps/tps tables
- dfco

ts-panels-2


mck1117:
set your display to 1360x768 and see if they fit OK

alrijleh:
DFCO just barely fit, I moved the curve to the side to make it less crowded.

The idle settings page is too tall, I can address those in a future PR (unless someone else gets to it first)

mi-hol:
@mck1117 would this PR fit with ongoing cleanups?

PR #494, opened 2024-09-21
Reorganize Idle Settings

alrijleh:
The current idle settings menu is too tall to fit on a 1300x700 display, and is cumbersome to work with as a result

I've reorganized it into open loop and closed loop menus:
- open loop only contains settings relevant to open loop IAC position, includeing the CLT multiplier curve
- closed loop contains both IAC and ignition settings
    - it might be nice to include live data of PID adders, current IAC position, and timing here, but that may get too crowded

idle menu


alrijleh:
Thoughts on merging this in?  The current idle page is too tall for my low resolution tuning laptop

Issue #504, opened 2024-10-30
Disable TPS Error check on Auto Calibration


dynfer:
Hello,
Suggestion to disable TPS error check during the auto calibration procedure. This would allow Toyota installs to use it.
Currently on real way is to calibrate the TPS manually.

mck1117:
what specific behavior do you see that you think is wrong? screenshot/vid?

dynfer:
The throttle just stops at random values due to the TPS error, its not moving smoothly from 0 to 100%.

rusefillc:
> The throttle just stops at random values due to the TPS error, its not moving smoothly from 0 to 100%.

vid? logs? what throttle is that?

dynfer:
@rusefillc I've mainly encountered it with Toyota throttles which require the special setting for the secondary sensor. It basically errors out if setting isn't set right preventing the calibration from finishing.

dynfer:
Solved in #514

dynfer:
Solved in #514

PR #511, opened 2024-11-13
Disable TPS error check on autocal

dynfer:
https://github.com/FOME-Tech/fome-fw/issues/504

### DIFF
--- a/firmware/controllers/actuators/electronic_throttle.cpp
+++ b/firmware/controllers/actuators/electronic_throttle.cpp
@@ -532,7 +532,8 @@ bool EtbController::checkStatus() {
 
 	bool shouldCheckSensorFunction = engine->module<SensorChecker>()->analogSensorsShouldWork();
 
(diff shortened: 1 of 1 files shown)


firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -532,7 +532,8 @@ bool EtbController::checkStatus() {
 
 	bool shouldCheckSensorFunction = engine->module<SensorChecker>()->analogSensorsShouldWork();
 
-	if (!m_isAutotune && shouldCheckSensorFunction) {
+	if (!m_isAutotune && shouldCheckSensorFunction && !m_isAutocal) {
+		// Added check to remove TPS errors during autocalibration https://github.com/FOME-Tech/fome-fw/issues/504
```
mck1117, line 536: you can just do `#504` instead of the full link, we know what repo it's for 🙂

mck1117:
this doesn't help, bug is completely different as discussed on discord

PR #514, opened 2024-11-14
rewrite etb autocal to support main loop

mck1117:
ETB no longer has its own thread, so we can't block execution with the calibration routine. Chop it up in to a state machine instead, using a timer instead of sleeps.


nmschulte:
This is solving a problem, yeah?  The TPS errors during autocal?

mck1117:
> This is solving a problem, yeah? The TPS errors during autocal?

It isn't exactly TPS errors during autocal. It's that running autocal completely blocks the ADC from getting read at all.

PR #517, opened 2024-12-05
Fix dynamic labels for GPPWM tables

dynfer:
This allows the GPPWM tables to change the axis name dynamically.


rusefillc:
How about some attribution?

dynfer:
> How about some attribution?

Thank You for the initial finding.

rusefillc:
> Thank You for the initial finding.

I think you should do a bit better, how about link original commit?

PR #520, opened 2024-12-09
logging central: memseting 6500 bytes is waste of time

rusefillc:
And avoiding a memset!


nmschulte (approved):

rusefillc:
@mck1117 what do you think?

PR #522, opened 2024-12-10
fix clang 19 warnings/errors

nmschulte:
Using clangd/clang 19 and Neovim LSP client/server, these changes were warnings (or sometimes errors).


rusefillc:
@nmschulte is there/should be there a CI configuration showing these warnings/errors?

nmschulte:
> @nmschulte is there/should be there a CI configuration showing these warnings/errors?

That is an interesting idea, but it would require running `clang`/`clangd` vs `gcc`.  I can look into it if it's something we're interested in.  To dial in my `clangd` configuration I learned about `clangd --check=FILE` which runs `clangd` for just one file.

https://clangd.llvm.org/guides/include-cleaner seems to be the feature that is telling about these `#include`s; I don't think GCC provides a similar mechanism.

> clangd has features to help you maintain the set of #include directives at the top of each file that are used to import dependencies.
> 
> It follows the [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md) model: each source file should #include headers that declare the symbols it references, and no others.
> 
> This means:
> 
>     files should not rely on transitive includes, only headers they include directly
>     #include directives describe direct dependencies between files, and it’s possible to reason about them locally
> 
> This is opinionated, and stricter than “will it compile”. It implies adding #includes that the compiler does not strictly need. Some #includes that satisfy the compiler should be replaced with more specific ones. If your codebase broadly aims to follow this style, you should enable these features.

Worth reiterating: ***This is opinionated***, though that isn't necessarily a bad thing: it's helpful.

rusefillc:
@nmschulte does not macos build run clang by default/always?

more or less my question is: why change anything, what's the problem statement?

nmschulte:
I don't believe MacOS uses clang; I believe it uses Arm's GCC for that platform.

I suppose in essence, what this PR is proposing is: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md

I wasn't quite aware of that when I created this PR, but I think it's worth considering.

rusefillc:
@nmschulte ``-Xiwyu: Tells GCC to use the IWYU plugin.``

nmschulte:
> -Xiwyu: Tells GCC to use the IWYU plugin.

@rusefillc where's that reference from?

rusefillc:
@nmschulte google AI could be hallucinating?

image

mck1117:
> I don't believe MacOS uses clang; I believe it uses Arm's GCC for that platform.

macos uses clang for unit tests

nmschulte:
capturing here as best place currently related to clang goodness:
https://www.phoronix.com/news/Linux-Clang-Thread-Safety

rusefillc:
@nmschulte rebase is needed?

PR #544, opened 2025-01-25
TS/MLV: Lambda / AFR Datalog entry

### LINKED ISSUE #543: VE Analyze broken
Hello,

After using some logs with 12.12.2024 firmware ve analyze is broken.

The problem stems from the fact we dont have an AFR or Lambda field anymore.
Currently we have Air/Fuel Ratio and Lambda1/2/3/4 which are not correctly picked up by MLV since it requires specific names ie. AFR or Lambda.

Currently a workaround is to link the fields in MLV.

nmschulte:
fixes #543

alias `Lambda` or `AFR` TS/MLV Datalog entry


mck1117:
picked a bit cleaner in 3089a05386e3c3b7229da80ba05870b19219e660

PR #546, opened 2025-01-25
remove gen_default_config.sh / gen_default_everything.sh

### LINKED ISSUE #538: remove gen_config_default.sh/gen_default_everything.sh
https://github.com/FOME-Tech/fome-fw/blob/master/firmware/gen_config.sh#L58

> ```sh
> #
> # TODO: it's time to kill the 'default' bundle concept and just live happily with explicit f407-discovery
> #
> # problem statement: it's desired that plain 'make' invocation compiles discovery binary
> # reality: while we have per-target signature*.h and rusefi*.ini we do not have matching per-target 'engine_configuration_generated_structures.h'
> # so for plain 'make' to produce f4 discovery we rely of file system engine_configuration_generated_structures.h being f4 discovery version
> #
> #
> # default config should be generated after normal custom boards so that it would be default
> # firmware/controllers/generated/rusefi_generated.h file which would be pushed into VCS
> ```

----

```
.github/workflows/gen-configs.yaml
44:        ./gen_config_default.sh

firmware/gen_config.sh
68:./gen_config_default.sh

firmware/gen_config_default.sh
2:# file gen_config_default.sh

firmware/gen_default_everything.sh
8:echo "gen_config_default.sh as part of everything"
9:./gen_config_default.sh
```
```
.github/workflows/build-FOME-console.yaml
30:        run: ./gen_default_everything.sh
```

nmschulte:
This just nukes the top-most layer of things; removes dependencies on these files, clarifies a few things, and beginnings of fixes #538


firmware/gen_config.sh
```diff
@@ -55,18 +49,4 @@ for BOARD in \
...
-# TODO: it's time to kill the 'default' bundle concept and just live happily with explicit f407-discovery
-#
-# problem statement: it's desired that plain 'make' invocation compiles discovery binary
-# reality: while we have per-target signature*.h and rusefi*.ini we do not have matching per-target 'engine_configuration_generated_structures.h'
-# so for plain 'make' to produce f4 discovery we rely of file system engine_configuration_generated_structures.h being f4 discovery version
-#
-#
-# default config should be generated after normal custom boards so that it would be default
-# firmware/controllers/generated/rusefi_generated.h file which would be pushed into VCS
-./gen_config_default.sh
```
nmschulte, line 68: I can't find an existing flow that relies on this invocation here; `gen-configs.yaml` already has this explicitly after the call to this `gen_config.sh` and is the only flow I could find, unless somehow the GH build containers are re-used on purpose in some scenario/way I'm unaware of.

firmware/gen_config.sh
```diff
@@ -8,13 +8,7 @@ echo "The storage section of fome.ini is updated as well"
 rm -f gen_config.log
 rm -f gen_config_board.log
 
-# todo: who is the consumer of this folder? shall we move that 'mkdir' command closer to usage?
-mkdir build
```
nmschulte, line 12: the Make build, ChibiOS's `BUILDDIR`, is this -- it's unnecessary

firmware/config/boards/common_make.sh
```diff
@@ -18,7 +18,8 @@ FW_DIR=$(readlink -f $FW_DIR)
 echo "FW dir is $FW_DIR"
 cd $FW_DIR
 
-mkdir -p .dep
+mkdir -p .dep # ChibiOS build's DEPDIR
+mkdir -p build # ChibiOS build's BUILDDIR
```
nmschulte, line 22: this is actually configurable via `BUILDDIR`, as is I think `DEPDIR`, but these features are hardcoded in a few places, mainly around the `compile.sh` layer atop Make

mck1117 (approved):

PR #564, opened 2025-03-02
add APEX board

TurboMarian:


mck1117:
what is an `APEX board`? let's chat on discord before we merge this.

firmware/config/boards/APEX/connectors/generated_outputs.h
```diff
@@ -0,0 +1,55 @@
+//DO NOT EDIT MANUALLY, let automation work hard.
```
mck1117, line 1: don't check in generated files

firmware/config/boards/APEX/connectors/generated_ts_name_by_pin.cpp
```diff
@@ -0,0 +1,48 @@
+//DO NOT EDIT MANUALLY, let automation work hard.
```
mck1117, line 1: don't check in generated files

firmware/config/boards/APEX/connectors/KRYPTON.pngZone.Identifier
mck1117, line 1: what is this file, and the others like it? looks like some garbage file that we didn't intend to include

PR #574, opened 2025-03-24
cherry-pick from RE da0eba081ebcc0ae40cb67cf140d23b6f6358587

mi-hol:


mck1117:
why do we commit generated files? can you justify why this cherry pick?

mi-hol:
> why do we commit generated files? 

this was unintentionally added, I didn't notice commit https://github.com/FOME-Tech/fome-fw/pull/574/commits/7efd719ed154b15061912d96e4e43015125e1a1e


> can you justify why this cherry pick?

improved error handling with info to fix root cause of error.
But actually it was just a first test if PRs are getting looked at currently because a number of PRs (that look sensible to me) have had no response (ie #539 , #494) and the JAVA_console is currently getting a lot of improvements in RE

Issue #591, opened 2025-05-14
lua API for storing data in backup SRAM


mck1117:
There are scenarios where you might want to store some user "settings" from Lua that survive a reboot. We have backup sram and many ECUs have a coin cell or permanent 12v supply, so those should work.

PR #592, opened 2025-05-14
implement persistent lua values

mck1117:
#591

### DIFF
--- a/firmware/controllers/lua/lua_hooks.cpp
+++ b/firmware/controllers/lua/lua_hooks.cpp
@@ -948,6 +948,10 @@ void configureRusefiLuaHooks(lua_State* l) {
 		doScheduleStopEngine();
 		return 0;
 	});
+
+	lua_register(l, "getPersistentValue", lua_getPersistentValue);
+	lua_register(l, "storePersistentValue", lua_storePersistentValue);
+
 #if EFI_SHAFT_POSITION_INPUT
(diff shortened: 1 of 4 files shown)


nmschulte (approved):

firmware/controllers/lua/lua_hooks.cpp
```diff
@@ -394,6 +395,45 @@ static int lua_setAirmass(lua_State* l) {
...
+
+	return *backupRam;
+}
+
+static int lua_getPersistentValue(lua_State* l) {
+	auto idx = luaL_checkinteger(l, 1);
+
+	auto backupRam = checkBackupSram(l);
+
+	if (idx < 0 || idx > efi::size(backupRam.LuaPersistentData)) {
```
nmschulte, line 413: perform 0- vs 1- based index translation for the Lua scripter, as we do for other Lua APIs

also, bug here about `idx == 64` (`efi::size(backupRam.LuaPersistentData)`)

```suggestion
	if (idx < 1 || idx > efi::size(backupRam.LuaPersistentData)) {
		luaL_error(l, "invalid backup ram index: %d", idx);
	}
	
	lua_pushnumber(l, backupRam.LuaPersistentData[idx-1]);
```

nmschulte:
I need to ...
- provide an analysis of current Lua API w/re: 0-based vs 1-based indexing
- align verbiage and document effects/expectations/assumptions of the feature, for the Wiki

Issue #597, opened 2025-05-19
Changes to CAN Wideband behaviour


diy-efi:
I think it might be beneficial to be able to select the type of can wideband after enabling.

I add the wideband controller as a default on my hardware.
Someone is wanting to use a aem-x as that's in their cabin, and they want to see the data as well.
If he connects his aem-x and there's also the internal one.
I don't think there logic to say mck wideband doesn't have a sensor connected, I have valid aem data, use that. 
So I'm not sure exactly what the ECU will use as afr/lambda

I think a good solution would be that rus wideband controller  menu option is renamed to CAN Wideband Config like the popup is.

On the CAN Wideband Config popup.
There should be a selection of the type [ AEM-X, FOME, ... ].
There shoud be the ability to select which physical CAN bus the wideband hardware is attached to.
Then the FOME ideband stuff can be grouped with a suitable name, this can be greyed out when the type is non-FOME.

![Image](https://github.com/user-attachments/assets/c7dbcbe3-76d1-4f2a-940c-4d94405b6190)

PR #598, opened 2025-05-20
selectable CAN wideband controller mode

mck1117:
#597 

- Different UI gets shown depending on wideband type
- Only listen to CAN frames from the selected type (instead of all supported at the same time)

image
image
image


diy-efi:
Much nicer and user-friendly.

Two small things.
image
I think the label would be better as "Wideband controller type"

Then the options [ Analogue, FOME - CAN, AEM-X Series - CAN ]

Also, whilst "FOMEInternal" is correct for my application, it might be a bit confusing to those on legacy hardware with an external wideband controller based on your module.

Issue #466, opened 2024-08-16
fome-fw doesnt compile on mac with m1 chip


D0kRay:
If I run any task from vscode for example 'Build Hellen NB1' it always fails after the bootloader compilation with the following error: 

    Creating blbuild/fome_bl.list
    
       text    data     bss     dec     hex filename
      18416      64  131008  149488   247f0 blbuild/fome_bl.elf
    
    Done
    ../../common_make.sh: line 48: 0xTEXT: value too great for base (error token is "0xTEXT") 

The following toolchain is used: arm-gnu-toolchain-11.3.rel1-darwin-x86_64-arm-none-eabi
And this is the commandline output:

     *  Executing task: bash ./compile_hellen-nb1.sh 
    
    Entering common_make.sh with board hellen/hellen-nb1 and CPU ARCH_STM32F4
    Board dir is /Users/macuser/Documents/VSCode/fome-fw/firmware/config/boards/hellen/hellen-nb1
    FW dir is /Users/macuser/Documents/VSCode/fome-fw/firmware
    Calling make for the main firmware...
    found /Users/macuser/Documents/VSCode/fome-fw/firmware/config/boards/hellen/hellen-nb1/connectors/generated_ts_name_by_pin.cpp 
    PROJECT_BOARD: hellen/hellen-nb1
    BOARD_DIR:     /Users/macuser/Documents/VSCode/fome-fw/firmware/config/boards/hellen/hellen-nb1
    PROJECT_CPU:   ARCH_STM32F4
    SHORT_BOARD_NAME: hellen-nb1
    CPU_HWLAYER:   ports/stm32/stm32f4
    CONFDIR:       ./hw_layer/ports/stm32/stm32f4/cfg
    LDSCRIPT:      ./hw_layer/ports/stm32/stm32f4/STM32F4.ld
    UNAME_SM:          Darwin arm64
    COMPILER_PLATFORM: arm-gnu-toolchain-11.3.rel1-darwin-x86_64-arm-none-eabi
    TRGT:              ./ext/build-tools/arm-gnu-toolchain-11.3.rel1-darwin-x86_64-arm-none-eabi/bin/arm-none-eabi-
    echo "#pragma once" > gitversion.h.gen
    printf '#include "fan_control.h"\n #include "fuel_pump.h"\n #include "gear_detector.h"\n #include "trip_odometer.h"\n' > engine_modules_generated.h.gen
[...]

PR #621, opened 2025-06-29
Fix ARM Mac build

victordiges:
Added a hex2dfu built binary for arm64 for Mac.
Updated hex2dfu submodule to RusEFI prod branch
Force the use of objcopy and objdump from arm-none-eabi toolchain in common_make.sh
Use darwin hex2dfu binary in Mac in common_make.sh

Fixes https://github.com/FOME-Tech/fome-fw/issues/466


firmware/config/boards/common_make.sh
```diff
@@ -35,8 +35,12 @@ fi
 
 if uname | grep "NT"; then
```
nmschulte, line 36: I wonder if we can tap/factor the platform detection logic in `firmware/use_arm_gcc.mk`, in the future.

nmschulte: 1) I wanted to create awareness about it; glad you'd seen it.
2) We'd have to factor something to make them common, but I'm not sure along which lines makes sense (entirely w/in Make, via vars? or via callable `.sh` script? or ??)
3) I don't expect it to be part of this changeset; unnecessary unless there's an obvious solution.

Issue #508, opened 2024-11-06
Virtual Inputs


dynfer:
Hello,

Suggestion to add virtual inputs and a menu in TS to allow for easier use of can input data without lua to make it more attractive to users without a coding background.

Ideal scenario would I have button on CAN and I want it to activate my AC relay. I set it, go to AC settings menu and set the button input as that virtual switch.
A simillar approach would be beneficial for sensor inputs for example Virtual Sensor #1 could be assigned to VSS / ambient temp etc.

Additionally if we could attach some logic to those it would be beneficial, for example Virtual Input #1 == true and Digital Input #1 == true => virtual input #2 == true

PR #622, opened 2025-06-29
CAN virtual inputs

mck1117:
#508

Add configurable fake input pins that read a bit from a CAN message instead of reading a real pin

image


dynfer:
lgtm I guess we can send it

PR #623, opened 2025-07-01
AC Pressure switch and after start delay

dynfer:


firmware/integration/rusefi_config.txt
```diff
@@ -1277,8 +1278,10 @@ float tChargeAirDecrLimit;Maximum allowed rate of decrease allowed for the estim
 
 	uint8_t autoscale dfcoDelay;Delay before cutting fuel. Set to 0 to cut immediately with no delay. May cause rumbles and pops out of your exhaust...;"sec", 0.1, 0, 0, 10, 1
 	uint8_t autoscale acDelay;Delay before engaging the AC compressor. Set to 0 to engage immediately with no delay. Use this to prevent bogging at idle when AC engages.;"sec", 0.1, 0, 0, 10, 1
+	uint8_t autoscale acStartDelay;Deley before the AC compressor engages after the start of the engine. Set 0 to engage immediately;"sec", 0.1, 0, 0, 255, 1
```
mck1117, line 1281: this maximum is impossible - it's a uint8_t scaled by 0.1, so the maximum can't be larger than 25.5 seconds

mck1117: (I went ahead and fixed it)

PR #628, opened 2025-07-03
use Chibi Contrib on all platforms for CRC

dynfer:


mck1117:
I think this is done by the commits linked from #631

Issue #627, opened 2025-07-02
Make: generated TS .ini does not depend tunerstudio.template.ini


nmschulte:
simply editing `firmware/tunerstudio/tunerstudio.template.ini` and running `make` (after a prior successful) does not result in updated `firmware/tunerstudio/generated/*.ini`.

mck1117:
does 37fa7d7 fix this issue for you?

nmschulte:
seems to have this behavior now.

PR #629, opened 2025-07-03
Make: generated TunerStudio .ini

nmschulte:
trying to tackle #627

this works; building after editing only `tunerstudio.template.ini` execs `gen_board_config.sh`, however this updates timestamps on some files that `all` then again depends on, and a third run will re-build the `OBJS` ...


firmware/fome_generated.mk
```diff
@@ -5,9 +5,14 @@ GENERATED := \
 # $(GENERATED_DIR)/log_fields_generated.h \
 # $(GENERATED_DIR)/output_lookup_generated.cpp \
 # $(GENERATED_DIR)/rusefi_generated.h \
-# $(GENERATED_DIR)/value_lookup_generated.cpp
+# $(GENERATED_DIR)/value_lookup_generated.cpp \
+# $(PIN_NAMES_FILE)
 # TODO: how do we list multiple dependencies without the build happening multiple times?
```
nmschulte, line 10: I'm not sure I understand "build running multiple times" ... is this the "two runs to stabilize Make" concept I've described, or different?

PR #632, opened 2025-07-06
Modular config txt

rusefillc:


mck1117:
what's the purpose of this change? what usability/cpu/memory advantage does it bring?

rusefillc:
this change contributes towards code base quality which indirectly benefits end users

key words encapsulation, god object

PR #633, opened 2025-07-07
add RunningToIdleTransition test

FDSoftware:
upstream issue: https://github.com/rusefi/rusefi/issues/8206
also relevant pr: https://github.com/rusefi/rusefi/pull/8218

### DIFF
--- a/firmware/controllers/actuators/idle_thread.h
+++ b/firmware/controllers/actuators/idle_thread.h
@@ -86,6 +86,12 @@ public:
 		return m_lastPhase == Phase::Idling || (engineConfiguration->useSeparateIdleTablesForCrankingTaper && m_lastPhase == Phase::CrankToIdleTaper);
 	}
 
+#ifdef EFI_UNIT_TEST
+	Pid *getIdlePid() {
+		return &m_pid;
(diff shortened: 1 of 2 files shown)


FDSoftware:
> @rusefillc @FDSoftware
> How do we reconcile the behavior in the test (expect zero integrator) with this comment?

This part in this PR, where I re-use that section of code to reset the integrator, but I still have doubts about how to refactor it, my idea was to modify as little as possible

https://github.com/rusefi/rusefi/pull/8218/files#diff-ca4992c714ac02b2480b9a5a1be662c6f8cc329ebdab31def07a1c0317e900c6R243-R246

This PR only contains tests showing the bug, because I didn't know if you were going to agree with the fix.

mck1117:
> This PR {633} only contains tests showing the bug, because I didn't know if you were going to agree with the fix.

I understand that - I'm debating whether it's a bug. The `FIXME` in this PR and the comment in the implementation (both in FOME and rusEFI!) disagree about the expected behavior.

In your (@rusefillc and @FDSoftware), under exactly which conditions should the idle PID be reset?

PR #644, opened 2025-07-23
oil pressure protection start delay

dynfer:


mck1117:
CI says something is wrong, can you look in to it?

firmware/controllers/limp_manager.cpp
```diff
@@ -127,29 +127,36 @@ void LimpManager::updateState(float rpm, efitick_t nowNt) {
...
-				isPressureSufficient = getOilSwitchState();
-			} else {
-				if (oilp) {
-					float minPressure = interpolate2d(
-											rpm,
-											config->minimumOilPressureBins,
-											config->minimumOilPressureValues
-										);
-					isPressureSufficient = oilp.Value > minPressure;
+			static bool afterStartDelayPassed = false;
```
mck1117, line 130: Stop putting `static` in class member functions! It's very bad for testability.

Issue #668, opened 2025-09-09
SensorN_PG Use this pin finally for something


NoTwixD:
As far as I know we don't use this pin for anything, we don't log its status, we don't check this pin before starting the engine or while driving?

usefull feature would be adding this status to logs , sometimes faulty sensor my shut down this tls115 chip and causes problem hard to track.

PR #693, opened 2025-10-20
check PG pins

dynfer:
#668


firmware/config/boards/proteus/board_configuration.cpp
```diff
@@ -128,6 +133,14 @@ static void setupSdCard() {
 	engineConfiguration->spi3mosiPin = Gpio::C12;
 }
 
+static void setupPowerGoodPins() {
```
mck1117: there's already a hook called `initBoardSensors()`, use that

DO NOT call `efiSetPad...` from `setBoardConfigOverrides()`

PR #695, opened 2025-10-23
trailing spark table

dynfer:
Switch trailing spark to use a table rather than scalar. Values still mean the same thing, degrees after the primary spark.

### DIFF
--- a/firmware/controllers/engine_controller.cpp
+++ b/firmware/controllers/engine_controller.cpp
@@ -518,8 +518,11 @@ bool validateConfig() {
 
 		ensureArrayIsAscending("Ignition load", config->ignitionLoadBins);
 		ensureArrayIsAscending("Ignition RPM", config->ignitionRpmBins);
-		ensureArrayIsAscending("Trailing spark load", config->trailingIgnitionLoadBins);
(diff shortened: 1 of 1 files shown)


firmware/integration/fome_config.txt
```diff
@@ -1582,6 +1580,9 @@ uint16_t[SCRIPT_TABLE_8] vvtTable2RpmBins;;"RPM", 1, 0, 0, 18000, 0
 int16_t[IGN_RPM_COUNT x IGN_LOAD_COUNT] autoscale ignitionTable;;"deg", 0.1, 0, -20, 90, 1
 uint16_t[IGN_LOAD_COUNT] ignitionLoadBins;;"Load", 1, 0, 0, 1000, 0
 uint16_t[IGN_RPM_COUNT] ignitionRpmBins;;"RPM",	   1, 0, 0, 18000, 0
+int16_t[IGN_RPM_COUNT x IGN_LOAD_COUNT] autoscale trailingIgnitionTable;;"deg", 0.1, 0, -20, 90, 1
```
mck1117: trailing spark angle can't be negative - it's an offset in degrees after the leading spark. This can probable be `uint8_t autoscale .... "deg", 0.1, 0, 0, 25, 1`

PR #709, opened 2025-12-03
more: remove kinetis, cypress hw layer ports, references

nmschulte:
#708 follow-on


mi-hol (commented):
@nmschulte that looks perfect!

PR #716, opened 2026-01-15
Add fuelFlow hook for lua

diy-efi:


diy-efi:
> cannot use
> 
> ```lua
> getOutput('fuelFlowRate')
> ```
> 
> ?
getOutput is only updated inside updateTunerStudioState()
So if your laptop is not connected, it doesn't work.

mck1117:
this is done in https://github.com/fome-tech/fome-fw/issues/721

PR #733, opened 2026-02-07
its not 2010 we have more threads

dynfer:


mck1117:
These are actually intentionally set low. Multiple actions runners exist on the same physical machine/vm, so if they all run at nproc * 1.5, 10 simultaneous builds will be 600 threads trying to run at the same time.

PR #738, opened 2026-03-19
further finalize debug_mode removal

nmschulte:
a few end-user bits of concern (TS ini)

also nix development notes; maybe overzealous however (DBG_LOGIC_ANALYZER; mazda VVT)


java_console/autotest/src/main/java/com/rusefi/f4discovery/PwmHardwareTest.java
```diff
@@ -57,6 +57,7 @@ public void scheduleBurnDoesNotAffectTriggerIssue2839() {
         assertEquals("totalTriggerErrorCounter", triggerErrors, totalTriggerErrorsNow);
     }
 
+    // TODO:: unignore; https://github.com/FOME-Tech/fome-fw/issues/731 (re-enable VssHardwareLoopTest, ...)
```
mck1117, line 60: why do we make unrelated changes in the same pr

nmschulte: IIUC, it is related by line ~78 of the test, which is why it is `@Ignore`'d -- added the comment/issue ref. to clarify things; perhaps the `unignore;` is redundant

PR #744, opened 2026-04-16
Add FOME CAN Bridge tool and firmware ISO-TP support

DFlak:


firmware/console/binary/serial_can.cpp
```diff
@@ -38,7 +38,7 @@ static CanTsListener listener;
 int CanStreamerState::sendFrame(
 		const IsoTpFrameHeader& header, const uint8_t* data, int num, can_sysinterval_t timeout) {
 	int dlc = 8; // standard 8 bytes
-	CanTxMessage txmsg(CAN_ECU_SERIAL_TX_ID, dlc, CanBusIndex::Bus0, false);
```
mck1117, line 41: please split the ECU part out to a separate PR

build-bridge.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
nmschulte, line 1: nit; prefer: `#!/usr/bin/env bash`

PR #745, opened 2026-04-16
Add Yellowbox (STM32H743) board support and bootloader stubs

DFlak:


firmware/bootloader/openblt_chibios/openblt_net.cpp
```diff
@@ -9,7 +9,7 @@ extern "C" {
 #include "net.h"
 }
 
-static ServerSocket server;
+static ServerSocket server("openblt");
```
mck1117, line 12: unrelated change?

firmware/config/boards/yellowbox/connectors/generated_ts_name_by_pin.cpp
```diff
@@ -0,0 +1,53 @@
+// DO NOT EDIT MANUALLY, let automation work hard.
```
mck1117, line 1: do not commit generated files

firmware/tunerstudio/generated/fome_yellowbox.ini
```diff
@@ -0,0 +1,8286 @@
+; This is a TunerStudio project for the FOME.tech engine management system
```
mck1117, line 1: do not commit generated files

mck1117:
The bar for adding a new board to mainline is very high because it increases the maintenance cost of the firmware. Any changes may have to be verified across a huge matrix of different board configs.

1. Do more than 10 ECUs exist in the wild using this firmware?
2. Commercial product or fun project?
3. Is there a good reason the next revision of this hardware can't use the Atlas firmware build? It looks like it's targeting  the same/similar chip as Atlas with a similar feature set. You will receive much better firmware support and CI builds if you target an existing firmware rather than create a new one.

firmware/util/crc_accelerator.cpp
```diff
@@ -83,8 +83,8 @@ bool checkFirmwareImageIntegrity(uintptr_t baseAddress) {
 	uint8_t* start = reinterpret_cast<uint8_t*>(baseAddress);
 	size_t imageSize = *reinterpret_cast<size_t*>(start + checksumOffset + 4);
 
-	if (imageSize > 1024 * 1024) {
-		// impossibly large size, invalid
+	if (imageSize > 2 * 1024 * 1024) {
```
mck1117, line 86: unrelated change?

PR #748, opened 2026-04-16
Refactored CAN-TS-Bridge ecu sided changes

DFlak:


firmware/console/binary/serial_can.h
```diff
@@ -120,19 +132,20 @@ class CanTsListener : public CanListener {
 	}
 
 protected:
+	uint32_t m_id = 0;
```
mck1117, line 135: this shadows the base class's m_id, intentional?

mck1117: (add a protected setter to set the base class member, rather than shadowing)

firmware/console/binary/serial_can.cpp
```diff
@@ -387,11 +350,12 @@ void CanTsListener::decodeFrame(const CANRxFrame& frame, efitick_t /*nowNt*/) {
...
 #if HAL_USE_CAN
 
-void CanStreamer::init() {
+void CanStreamer::init(uint32_t rxId) {
+	listener.init(rxId);
 	registerCanListener(listener);
 }
 
-can_msg_t CanStreamer::transmit(canmbx_t /*mailbox*/, const CanTxMessage* /*ctfp*/, can_sysinterval_t /*timeout*/) {
+can_msg_t CanStreamer::transmit(uint32_t /*eid*/, canmbx_t /*mailbox*/, const CanTxMessage* /*ctfp*/, can_sysinterval_t /*timeout*/) {
```
mck1117, line 358: why add this parameter? it's never used!

mck1117:
full claude review (don't necessarily have to do it all): https://gist.github.com/mck1117/c688becc107424a46cc9101d3e1c6dc8

PR #758, opened 2026-04-25
Updated Viper trigger name

twisted-builds-llc:
All this pull request does is update the viper trigger pattern name to 1996-2006 Dodge Viper so the end user knows this trigger configuration is only for those years of engines.


twisted-builds-llc:
Closed pull request due to branch conflict.

PR #759, opened 2026-04-26
Update viper wheel definition and correct wasted spark for viper.

twisted-builds-llc:
This pull request does two things.

1st. Updates the fome_config and changelog to indicate the Viper wheel definition added is for 1996-2006 Dodge Vipers.

2nd. Due to Dodge Vipers ignition system (wasted spark coil packs fired sequentially) pin_respiratory.cpp was changed to where if, and only if, the 1996-2006 Dodge Viper wheel pattern is selected, it will allow the end user to setup 1 ignition coil output to two different cylinders - allowing proper sequential odd fire "wasted spark" to function correctly.


firmware/hw_layer/pin_repository.cpp
```diff
@@ -54,13 +54,31 @@ int brainPin_to_index(brain_pin_e brainPin) {
...
+static bool viperCoilPinShareAllowed(const char* existingUser, const char* requestedUser) {
+	if (existingUser == nullptr || requestedUser == nullptr) {
+		return false;
+	}
+
+	if (engineConfiguration->trigger.type != trigger_type_e::TT_VIPER_V10_CRANK) {
+		return false;
+	}
+
+	return strncmp(existingUser, "Coil ", 5) == 0 && strncmp(requestedUser, "Coil ", 5) == 0;
```
mck1117, line 69: Checking based on the coil string name isn't gonna fly. There's logic to make sure coil pins don't get hung on (overdwell), and this will probably break that.

Why do we need to share one pin for two outputs? What is the expected output pattern on a viper that we don't support already?

mck1117:
I think it's done in aa81deb3c6

PR #761, opened 2026-05-04
fix sdcard hangs bl

dynfer:
It was discovered recently that when a SDcard is inserted into an Atlas board and the user attemps to perform an OpenBLT update via fome console it leads to a kernel panic in ChibiOS.

After further investigation I've determined that the cause of this panic is related specifically to calling sdcStart() as second time while sdcp1 is already in BLK_READY state which violates the assert and results in a panic.

Ive attached the call stack
obraz

The fix I've found simply ommits the start procedure if the device has been already started/initialized.

### DIFF
--- a/firmware/hw_layer/mmc_card_attach.cpp
+++ b/firmware/hw_layer/mmc_card_attach.cpp
@@ -90,6 +90,10 @@ BaseBlockDevice* initializeMmcBlockDevice() {
 		return nullptr;
 	}
 
+	if (EFI_SDC_DEVICE.state == BLK_READY) {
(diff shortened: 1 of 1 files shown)


mck1117:
1. PR needs a description. Why this change? When is it necessary?
2. Code could use a comment. It's not obvious why/how this happens and why we should just return.

dynfer:
> 1. PR needs a description. Why this change? When is it necessary?
> 
>     2. Code could use a comment. It's not obvious why/how this happens and why we should just return.

done

PR #762, opened 2026-05-04
small fix to openblt update speed

dynfer:
FOME console writes firmware in 200 byte chunks here: OpenBltFlasher.java (line 109). But Atlas/OpenBLT advertises a max programming CTO of 200, and PROGRAM_MAX payload is maxProgCto - 1, so only 199 data bytes: XcpLoader.java (line 164).

That means every 200-byte console chunk becomes:

SET_MTA
PROGRAM_MAX with 199 bytes
PROGRAM with the remaining 1 byte

### DIFF
--- a/firmware/hw_layer/openblt/blt_conf.h
+++ b/firmware/hw_layer/openblt/blt_conf.h
@@ -153,9 +153,9 @@
 /** \brief Configure the desired communication speed. */
 #define BOOT_COM_RS232_BAUDRATE          (115200)
 /** \brief Configure number of bytes in the target->host data packet. */
-#define BOOT_COM_RS232_TX_MAX_DATA       (200)
+#define BOOT_COM_RS232_TX_MAX_DATA       (240)
 /** \brief Configure number of bytes in the host->target data packet. */
-#define BOOT_COM_RS232_RX_MAX_DATA       (200)
+#define BOOT_COM_RS232_RX_MAX_DATA       (240)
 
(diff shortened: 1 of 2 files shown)


mck1117:
why doesn't it use the advertised cto instead of hardcoding it

dynfer:
> why doesn't it use the advertised cto instead of hardcoding it


It does use the advertised CTO for the actual XCP packets.

The advertised value comes back from PROGRAM_START and gets stored in mTargetInfo.maxProgCto here: 
XcpLoader.java (line 108)
. Then writeData() uses that value to split the payload into PROGRAM_MAX or PROGRAM packets here: 
XcpLoader.java (line 157)
.

The “hardcoded” part you’re seeing in the flasher is one layer above that. In 
OpenBltFlasher.java (line 108)
, forEachFirmwareChunk(PROGRAM_CHUNK_SIZE, ...) is just deciding how much data to hand to writeData() at a time. That affects:

how often we do SET_MTA
progress update cadence
how large each host-side batch is
It is not the final wire packet size. writeData() still re-chunks that batch using the target’s advertised maxProgCto.

So the short answer is:

XcpLoader already honors the advertised CTO
OpenBltFlasher still has an extra host-side batching heuristic on top

dynfer:
WIth the changes in the new commit I have succesfully cut the update time by nearly 50% from the inital 5.5m to 1.5min without any major adjustments.

PR #768, opened 2026-05-11
G0 GPIO expansion with ADC and outputs

dynfer:
The implementations covers the following features of the G0

- ADC Inputs with 4x hw oversampling at 160samples with software averaging
- Outputs with the ability to work in 0/1 or PWM mode
- Digital Inputs Currently only 0/1 mode, frequency input and SENT is currently under development

The G0 extension is additionally managed via the main MCU, it keeps the current fw for the G0 in the form a header and if necessary it updates it, on top of that it has the option to flash a freshly installed G0 via the builtin ST bootloader.

[proteus-G0_Extension.pdf](https://github.com/user-attachments/files/27598734/proteus-G0_Extension.pdf)


firmware/hw_layer/g0_extension/g0_extension_output_manager.cpp
```diff
@@ -0,0 +1,92 @@
+#include "g0_extension_io_impl.h"
+
+#if EFI_PROD_CODE
+
+namespace g0_extension {
+namespace {
+
+static void putU16(uint8_t* buffer, uint8_t offset, uint16_t value) {
```
mck1117: why manually parcel/unparcel? after all, both ends are the same endianness and guaranteed to be in sync, so we can just share a header with structs like the wideband does

dynfer: done

PR #772, opened 2026-05-14
Enable MAX31855 on H7

dynfer:
The current implementation for MAX31855 was written way before H7, hence its lacking support for it in the current form. This PR enable its use on atlas.


firmware/hw_layer/sensors/max31855.cpp
```diff
@@ -144,12 +144,15 @@ static void egtRead() {
 }
 
 void initMax31855(spi_device_e device, egt_cs_array_t max31855_cs) {
+#if defined(HW_ATLAS) || (defined(HW_PROTEUS) && defined(STM32H7XX))
+	driver = getSpiDevice(SPI_DEVICE_5); // H7 has only SPI5 available for aux peripherals
```
mck1117, line 148: is this true? what about a proteus/polygonus h7? what about some other future H7 board that we're going to have to debug why this doesn't work?

java_tools/configuration_definition/src/main/java/com/rusefi/ConfigDefinition.java
```diff
@@ -181,6 +181,16 @@ public static void doJob(String[] args, ReaderStateImpl state) throws IOExceptio
             System.out.println("Signature: " + signature);
         }
 
+        {
+            // MAX31855 is fixed to SPI5 on H7, so the TS device selector should be hidden.
```
mck1117: noooooo do not do this, just set an override in the board_configuration.cpp

firmware/hw_layer/sensors/max31855.cpp
dynfer: Checked the schematics and both atlas and proteus h7 have spi5 as the aux spi.

Future boards will fallback to the selector.

mck1117:
picked as 91204ba and faf33b8

PR #773, opened 2026-05-15
Improve error logging for incompatible firmware format

mi-hol:
- Display and log identical error message.
- Added a recommended fix message for incompatible firmware format.


mi-hol:
@mck1117 I wonder if you noticed the pending PR?

PR #774, opened 2026-05-17
H7max31855 v2

dynfer:
followup to #772 

This PR intends to improve the current use of SPI peripherals to allow multiple SPI devices to function freely on the same SPI bus. The current implementation frequently starts and stops the peripheral with its own spi config. This change allows it to function with other devices on the same bus via spiAcquireBus / spiReleaseBus


dynfer:
will further handled in #778

PR #785, opened 2026-07-21
Add roadside and workshop flood clear modes

gdiciocco:
## Summary

- add selectable Roadside and Service / Workshop flood-clear modes
- keep the existing Roadside behavior, which restores fuel once running RPM is reached
- latch the Workshop fuel cut through accidental combustion until the accelerator is released
- skip the priming pulse using `DriverThrottleIntent`, including DBW applications
- expose the mode in TunerStudio without changing the persistent configuration layout

## Motivation

Flood clear was conditioned directly on `!isRunning()`. Residual fuel could therefore produce an accidental combustion event, push the engine over the cranking RPM threshold, and immediately re-enable injection.

Roadside mode intentionally preserves that behavior so the engine may recover and start. Service / Workshop mode keeps injection disabled after it has been armed, allowing cylinders to be cleared or a racing engine to be cranked for oil pressure. Releasing the accelerator clears the latch, and the mode cannot arm on an already-running engine.

The new setting occupies the previously unused bit 31 at offset 1064, so no existing fields move and the configuration size is unchanged. Roadside remains the default.

## Validation

- `make -j4` in `unit_tests`
- `./build/fome_test '--gtest_filter=limp.*:priming.*'`
- 12/12 selected tests passed

### DIFF
--- a/firmware/controllers/limp_manager.cpp
+++ b/firmware/controllers/limp_manager.cpp
@@ -261,9 +261,32 @@ void LimpManager::updateCutsInjectorDuty(float rpm, efitick_t nowNt, Clearable& 
 }
 
 void LimpManager::updateCutsFloodClear(Clearable& allowFuel) {
-	// If the pedal is pushed while not running, cut fuel to clear a flood condition.
-	if (!engine->rpmCalculator.isRunning() && engineConfiguration->isCylinderCleanupEnabled &&
-		Sensor::getOrZero(SensorType::DriverThrottleIntent) > CLEANUP_MODE_TPS) {
+	bool floodClearRequested = engineConfiguration->isCylinderCleanupEnabled &&
+		Sensor::getOrZero(SensorType::DriverThrottleIntent) > CLEANUP_MODE_TPS;
+
+	// Releasing the pedal (or disabling flood clear) always releases the workshop latch.
+	if (!floodClearRequested) {
+		m_floodClearLatched = false;
+		return;
+	}
+
+	if (!engineConfiguration->isCylinderCleanupLatching) {
+		// Roadside mode intentionally restores fuel if accidental combustion pushes
+		// the engine above the cranking RPM threshold.
+		m_floodClearLatched = false;
+		if (!engine->rpmCalculator.isRunning()) {
+			allowFuel.clear(ClearReason::FloodClear);
+		}
+		return;
+	}
+
+	// Workshop mode may also be used to crank a racing engine for oil pressure.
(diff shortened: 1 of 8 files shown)


firmware/controllers/limp_manager.cpp
```diff
@@ -261,9 +261,32 @@ void LimpManager::updateCutsInjectorDuty(float rpm, efitick_t nowNt, Clearable&
...
+	}
+
+	// Workshop mode may also be used to crank a racing engine for oil pressure.
+	// Arm only while stopped/cranking so selecting this mode cannot cut a running engine.
+	if (!engine->rpmCalculator.isRunning()) {
+		m_floodClearLatched = true;
+	}
+
+	if (m_floodClearLatched) {
 		allowFuel.clear(ClearReason::FloodClear);
```
mck1117, line 290: why not just something like 
```c++
if (engineConfiguration->floodClearCutsIgn) {
    allowIgnition.clear(ClearReason::FloodClear);
}
```
so that the engine will never start if that's set?

gdiciocco: I agree that cutting ignition in the latched / Workshop mode is something I should have included in the original PR.

I also appreciate the reduced complexity of your proposal and the more technical framing provided by a setting such as floodClearCutsIgn.

My concern with removing the fuel latch entirely is that, in some workshop situations, the engine may exceed the configured cranking threshold without commanded ignition. Examples include cranking with the spark plugs removed, an externally driven engine on a test bench, hot-spot autoignition, a slightly higher starter speed because the coils are no longer being charged or because the engine is being powered from an external battery booster in a workshop oil priming scenario after engine reassembly.

Once the ECU classifies the engine as running, the existing !isRunning() condition becomes false and both flood-clear cuts are removed. Injection and ignition are therefore restored simultaneously while the throttle is still open. Since the running state remains latched until the engine is detected as stopped, even a transient threshold crossing can create a real opportunity for the engine to start.

The ignition cut prevents the usual spark-driven threshold crossing, but it does not preserve the Workshop invariant if the threshold is exceeded for another reason. For that reason, I see ignition cut as a necessary addition to Workshop mode, but not as a complete replacement for the fuel latch.

My preference would be:
Roadside: existing behaviour, allowing injection to resume at running RPM.
Workshop: latch both fuel and ignition cuts until the accelerator is released.

There is also the separate priming-pulse path. On DBW applications a prime could still occur unless the skip condition uses DriverThrottleIntent, so I think that part of the PR should remain.

That said, I understand the value of the simpler implementation and I am happy to adjust the PR based on which guarantee we want Workshop mode to provide.

PR #786, opened 2026-07-31
Make Hella PULS oil level sensor usable

kjngineering:
## What this does

Makes the Hella PULS ultrasonic oil level sensor usable. The decoder has existed since 89945206 ("stub hella level sensor", Sept 2023) but was never reachable: `HellaOilLevelSensor` is compiled via `sensors.mk` yet never instantiated, there is no `OilLevel` sensor type, no configuration field, and no way to assign it a pin. This wires it up and corrects the decode.

Tested against a `6PR 009 622-041` on the bench (Datsun 260Z / VH41DE project), cross-checked against the Hella datasheet and a known-working Arduino decoder for the same part.

## Changes

**Wiring up**
- New `SensorType::OilLevel`, `oilLevel` output channel, and a TunerStudio dialog under Sensors with input pin and variant selection
- `init_hella_oil_level.cpp` hooked into `reconfigureSensors()` / `stopSensors()`
- `deInit()` so the sensor survives reconfiguration

**Decode corrections**
- Level scale was `0mm at 23ms -> 150mm at 87.86ms`, which matches no shipping variant. Correct values come from the datasheet transfer function `Level = [(T2/T - 0.2) x k + offset] x T/110 + 5.34`, which resolves to:
  - `6PR 009 622-041` (5v, 45mm range): **20.9mm to 65.9mm**
  - `6PR 009 622-051` (12v, 129mm range): **18.0mm to 147.0mm**
  
  Selected by a new `hellaOilLevelVariant` config field rather than hardcoded, since the two parts differ only in scaling.
- Temperature was computed and then discarded (`setValidValue` commented out). Now published to `SensorType::OilTemperature`, `-40C` to `160C`.
- Data endpoints moved from 23/87ms to **22/88ms**. The datasheet defines 20% (22ms) and 80% (88ms) as the ends of the measuring range, not fault codes. The old guards rejected anything below 22.8ms or above 87.2ms, so legitimate full-scale readings were being thrown away — and with the datasheet's ±10% tolerance on all time signals, a nominal 22ms pulse can arrive anywhere from 19.8 to 24.2ms, which put the 22.8ms threshold inside the valid band.

Sync timing (780ms rising-to-rising break, 110ms inter-pulse) is unchanged — it was already correct. For anyone checking against the datasheet, 780ms is the 68.2ms diagnostic pulse plus the 711.8ms idle remainder of the 1000ms frame.

## Two latent bugs fixed along the way

- **Duplicate registration was a hard fault.** `OilTemperature` is already registered by `init_thermistors.cpp` when an analog sender is configured, and `SensorRegistryEntry::Register()` raises a `firmwareError` on a second registration. Anyone running both would have faulted the ECU on boot. The Hella temperature channel now checks `hasSensor()` first and yields to the analog sender, which keeps priority.
- **`unregister()` does not check ownership** — it clears the slot unconditionally. Tearing down the Hella sensor would therefore have evicted an analog sender's `OilTemperature` registration. Now tracked explicitly, and `deInit()` returns early if the sensor was never initialized.

## Testing

`unit_tests/tests/sensor/test_hella_oil_level.cpp` — 7 tests driving synthetic pulse trains through the two-argument `onEdge(nowNt, value)` overload:

- level scaling at both ends and midpoint, for both variants
- temperature scaling at both ends and midpoint
- level and temperature are not confused with one another
- an implausible pulse width drops sync and does not update the reading
- signal loss invalidates via the 2 second timeout rather than latching
- an analog oil temperature sender keeps ownership, and teardown does not evict it

## Notes for review

- Generated files are deliberately not included, per the note in `fome_config.txt` and the `gen-configs` workflow. This means `auto_generated_sensor.h` is missing a case for the new enum value until that workflow runs, which produces a `-Wswitch` warning; it is not fatal, and name lookup (`findSensorByName`) will return `"unknown"` for `OilLevel` until regeneration.
- `fome_config.txt` has no `mainUnusedEnd[]` padding to consume, so the two new fields are appended at the end of `engine_configuration_s` where they do not shift the offset of any existing field.
- **On atomicity:** CONTRIBUTING.md asks for one PR per distinct change, and this arguably bundles a feature with two bug fixes. I have kept it together because the code is currently unreachable — there is no working configuration to split toward, and the scale fix and temperature channel are only observable once the sensor can be assigned a pin at all. Happy to break it up if you would prefer.

### DIFF
--- a/firmware/CHANGELOG.md
+++ b/firmware/CHANGELOG.md
@@ -37,6 +37,7 @@ or
  - IMU data (lateral/longitudinal acceleration and yaw rate) from the BMW E90 MK60e1/MK60e5 DSC/ABS module
  - New VVT mode "Honda K24Z Exhaust" for the three-tooth unevenly spaced exhaust cam wheel used on the K24Z, which differs from the 4+1 wheel of earlier K series engines. This pattern is used by K series engines that use a 60-2 crank pattern.
  - CAN wideband controllers now report "time since last CAN frame received" and a TunerStudio alive indicator for all 4 wideband channels (previously only 2 had live data at all), so a dead/disconnected controller can be spotted even when its lambda reading is invalid
+ - Support for Hella PULS ultrasonic oil level sensors (6PR 009 622-041 and 6PR 009 622-051). A decoder for these sensors has existed since 2023 but was never configurable or connected to a sensor channel - it can now be assigned an input pin in TunerStudio and reports both oil level and oil temperature. Oil temperature is only published if no analog oil temperature sender is configured, which continues to take priority.
 
 ### Fixed
  - STM32F7 dual-bank ECUs no longer stall (potentially stopping the engine) when burning configuration with the engine running - configuration is now committed to flash when the engine is stopped #776
--- a/firmware/console/binary/output_channels.txt
+++ b/firmware/console/binary/output_channels.txt
@@ -293,4 +293,6 @@ struct_no_prefix output_channels_s
 
 	uint16_t[12 iterate] cylinderRpm;;"rpm", 1, 0, 0, 0, 0
 	int8_t[12 iterate] cylinderRpmDelta;;"rpm", 1, 0, 0, 0, 0
+
+	uint16_t autoscale oilLevel;@@GAUGE_NAME_OIL_LEVEL@@;"mm", 0.1, 0, 0, 0, 1
 end_struct
--- a/firmware/console/status_loop.cpp
+++ b/firmware/console/status_loop.cpp
@@ -375,6 +375,8 @@ static void updateMiscSensors() {
 	engine->outputChannels.auxSpeed1 = Sensor::getOrZero(SensorType::AuxSpeed1);
 	engine->outputChannels.auxSpeed2 = Sensor::getOrZero(SensorType::AuxSpeed2);
 
+	engine->outputChannels.oilLevel = Sensor::getOrZero(SensorType::OilLevel);
+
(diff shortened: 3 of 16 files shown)


kjngineering:
Closing for now - I want to complete bench testing against real hardware before this gets reviewer time. Will reopen once validated.

mck1117:
> Closing for now - I want to complete bench testing against real hardware before this gets reviewer time. Will reopen once validated.

you can also mark it as a draft rather than close, if you'd like

kjngineering:
Noted. I was actually considering some scope increase here, but want your opinion first @mck1117 .

Typically oil temperature is measured at the engine entry point (post pump, post cooler) - however, there are valid measurement points before this - e.g. (sump, tank, pre-cooler etc).

1) Can we add another SensorType for the above measurement? if so, what should it be called?
2) In this PR should I increase the OilLevel scope to accept voltage (e.g. 0-5V capacitive) or resistive (AT input - Float) type sensors - or is this better served in a separate PR?

mck1117:
1. I'd sort of be inclined to just call it "oil temperature 2". Not sure we need to be super prescriptive about where your two sensors are. FWIW, I've run single sensors in all those places before (pan vs. main feed gallery) as the only oil temp sensor. Another alternative is to use one of the aux temp inputs for the second sensor and let the one mapped as `OilTemperature` be the primary.
2. Does such a sensor actually exist? I know they do for fuel tanks, not sure if anybody uses those for oil.

I would split those in to separate PR(s) if/when we do them.

PR #797, opened 2026-08-29
Minor optimization to clone heavy submodules

alex-aparin:
Cloning submodules takes long time for me especially chibios cloning. This PR optimizes a little bit cloning by restricting history depth of submodules.


mck1117:
Meh. Full clones are rare. I'm not sure it's worth the added complexity.

PR #805, opened 2026-09-09
Add support for the Hella OPS+T oil pressure and temperature sensor

gdiciocco:
## Summary

Add native support for the Hella OPS+T combined digital oil pressure
and temperature sensor.

> [!NOTE]
> This FOME port has not yet been tested on hardware. The PR will remain
> in draft until both bench validation and testing on a running engine are
> complete.

The driver decodes the sensor's three-symbol PWM frame and publishes
its readings through the existing `SensorType::OilPressure` and
`SensorType::OilTemperature` channels. Existing consumers such as
logging, CAN, gauges, and protection logic therefore require no
sensor-specific changes.

## Implementation

- Add a configurable `hellaOpsTInput` using the existing EXTI input
  infrastructure.
- Decode the diagnosis/sync, temperature, and absolute-pressure
  symbols from both signal edges.
- Normalize pulse widths against the measured symbol period to
  compensate for the sensor oscillator tolerance.
- Convert absolute oil pressure to the gauge pressure expected by
  FOME using the current BARO reading, with a 101.325 kPa fallback.
- Clamp gauge pressure to zero when absolute pressure is below BARO.
- Use the existing `StoredValueSensor` timeout handling.
- Reset and resynchronize the decoder after malformed or unexpected
  symbols.
- Preserve existing oil pressure and temperature providers independently:
  OPS+T supplies only channels that are not already registered by another
  provider.
- Avoid acquiring an EXTI resource when both channels already have a
  provider.
- Keep the decoder independent from persistent configuration; the global
  instance and configuration binding live in `init/sensor`.

`efiExtiEnablePin()` now reports whether the EXTI resource was
successfully acquired. This lets the OPS+T driver roll back its sensor
registrations without later disabling an EXTI channel owned by another
input. Existing callers may ignore the returned value.

## Diagnostics

The OPS+T self-test distinguishes pressure, temperature, and hardware
faults.

Reported faults invalidate the affected sensor with
`UnexpectedCode::Inconsistent` and map to the existing range/performance
DTCs:

- Oil pressure: P0521
- Oil temperature: P0196

Existing timeout, low-input, and high-input mappings remain unchanged.

## Configuration

A single Hella OPS+T input field is added to the oil pressure settings
dialog in TunerStudio.

Only the configuration sources are included in this PR. Generated INI
files are intentionally left to the normal FOME generation workflow.

## Protocol reference

The protocol timing, scaling, diagnostic states, and oscillator tolerance
come from the official Hella brief information for part number
`6PR 010 378-207`:

- [Hella OPS+T oil pressure and temperature sensor datasheet](https://www.hella.com/forvia-us/assets/documents/BI_Oil_Pressure_and_Temperature_Sensor_2026.pdf)

The implementation was also cross-checked against
`gdiciocco/speeduino`, branch `caponord-stm32-optimized`, at commit
`14e36a962f952c753a6ec3fb3f935b4173bbfdb1`.

That Speeduino implementation has been tested on real hardware. This FOME
port has not yet been tested on real hardware, so the PR will remain a draft
until it has passed both bench validation and testing on a running engine.

## Testing

- OPS+T and oil DTC tests: 10/10 passed
- Full unit-test suite: 725/725 passed
- Full configuration generation: passed
- Core8 firmware build: passed
- `git diff --check`: passed

Tests cover:

- nominal temperature and pressure decoding;
- oscillator tolerance compensation;
- pressure, temperature, and hardware diagnostics;
- invalid pulse rejection and resynchronization;
- data received without a sync symbol;
- BARO fallback and zero-pressure clamping;
- timeout and subsequent recovery;
- per-channel existing-provider priority;
- deinitialization and reinitialization;
- P0521/P0196 diagnostic mapping.


mck1117:
have you seen https://github.com/FOME-Tech/fome-fw/pull/786? and that there's already an untested/partial implementation checked in?

gdiciocco:
Following the pattern discussed in #786, I made two cleanup changes before
hardware validation:

- Moved the global OPS+T instance and its `engineConfiguration` binding to a
  dedicated `init/sensor/init_hella_opst.cpp`. This keeps the decoder focused on
  the sensor protocol and independent from persistent configuration.
- Changed channel priority from inspecting the two ADC configuration fields to
  checking `Sensor::hasSensor()` for oil pressure and oil temperature. This
  preserves whichever provider is actually registered, avoids coupling OPS+T
  to the current analog implementations, and will also work with future
  non-ADC providers.

The driver still tracks ownership independently for both channels, so teardown
only unregisters channels acquired by OPS+T. If both channels already have a
provider, it does not acquire the EXTI input. The PWM decoding and conversion
behavior are unchanged by this refactoring. I also added the changelog entry
and the official Hella datasheet reference for part `6PR 010 378-207`.

The PR will remain in draft until the FOME implementation has passed both a
bench test and a test on a running engine.

@mck1117, one UI question before this leaves draft: the OPS+T input currently
lives in the existing Oil Pressure dialog to keep the configuration change
minimal, although the sensor supplies both oil pressure and oil temperature.

Would you prefer a dedicated Hella OPS+T dialog under Sensors, similar to the
approach proposed for the Hella PULS sensor in #786, or should we keep the
single field in the existing Oil Pressure dialog?

mck1117:
There is still duplicated code. Please do not create a second decoder for the same sensor.

edit: oh, this is a different sensor, not the oil level one. Is there any commonality between the oil level/temp sensor and this oil pressure/temp sensor we can extract? Any similarity in the pulse format?

gdiciocco:
@mck1117, could you clarify which duplicated decoder you are referring to?

The existing `HellaOilLevelSensor`, discussed in #786, appears to target the Hella PULS
`6PR 009 622-041/-051`, which reports oil level and temperature using a
roughly one-second frame with 110 ms slots and a 780 ms synchronization
interval.

This PR targets the Hella OPS+T `6PR 010 378-207`, which reports absolute oil
pressure and temperature. According to the
[Hella datasheet](https://www.hella.com/forvia-us/assets/documents/BI_Oil_Pressure_and_Temperature_Sensor_2026.pdf), it uses a
9216 us frame consisting of a 1024 us diagnostic symbol followed by 4096 us
temperature and pressure symbols. The measurement scaling and diagnostic
encoding also differ from the PULS protocol.

The two implementations do share the dual-edge PWM acquisition mechanism:
rising-to-rising period measurement, high-pulse measurement, basic pulse
validation, and capture reset. There does not appear to be an existing generic
period-and-duty capture helper in the current tree.

Did you mean that this common acquisition logic should be extracted and shared
by `HellaOilLevelSensor` and `HellaOpsTSensor`, while keeping their different
protocol state machines separate? Or is there another existing OPS+T decoder
that I have missed?

mck1117:
Yeah I saw "hella" and "pulse width" near each other and assumed the other sensor. Is there a real use case for this sensor that already exists? Can you show photos of a FOME ECU and the sensor in the same frame? I would prefer not to add drivers for obscure sensors that nobody will use.

Aside: Am I talking to an LLM or a human? I strongly suspect LLM or [meat proxying](https://gruhn.me/blog/2026-08-03/). I would prefer to talk to humans.

gdiciocco:
I personally can't tell which vehicles are equipped with this sensor. I chose this sensor as a way to have a reliable digital readout of both oil temperature and pressure from just the single oil bulb access, it required re-threading but it is a simple job. It can fit many engines natively, by rethreading, or with an adapter. Way more use cases than the PULS. I will also try to let it read CLT and pressure even if not deliberately sold for coolant. For me it is an outstanding sensor even in motorsport, for less than 40 eur digitally and reliably provides oil temperature and pressure where a similar bosch one, dual analog, costs over 150.

Here the LLM built list of vehicle equipping it:

- BMW: 1 Series F20/F21; 2 Series Coupé F22 and Convertible F23; 3 Series F30/F31 and GT F34; 4 Series F32/F33/F36; 5 Series F10/F11 and GT F07; X1 E84; X3 F25; X4 F26; X5 F15; Z4 E89. Mostly 1.6/2.0-litre turbo petrol engines, approximately 2010–2021. [Application catalogue](https://www.partvendo.com/sensor-yag-isisi-basinci-1096605)
- Volvo: S60 II (134), S80 II (124), V40 (525), V40 Cross Country (526), V60 I (155), V70 III (135), XC60 I (156), XC70 II (136), XC90 II (256), S90 II (234), and V90 II (235). The listed applications mainly cover 1.5/2.0-litre Drive‑E petrol and diesel engines, roughly 2013–2022. [Volvo model/engine list](https://spareto.co.uk/products/hella-sender-unit-oil-temperature-pressure/6pp-010-378-207)
- Aston Martin: DB11, DBR22, DBS, DBS Zagato, Valour, Vanquish, Vantage, and Vantage SP, under OE reference HY53-9S306-AA. The DB11 application is also supported by an Aston Martin technical document. [Aston Martin parts listing](https://astonmartinbits.com/products/hy53-9s306-aa), [DB11 technical document](https://static.nhtsa.gov/odi/tsbs/2017/MC-10108958-9999.pdf)
Cross-reference OE numbers: BMW 12 61 7 595 724 / 12 61 8 613 333, Volvo 31336469 / 31405635, Aston Martin/Ford HY53-9S306-AA. [Cross-reference source](https://www.motointegrator.at/artikel/2005932-sensor-oeltemperatur-druck-hella-6pp-010-378-207)

Actually I'm working on a second ECU on the bench, the sensor is installed in the motorbike and running from some years.

WhatsApp Image 2026-09-09 at 23 01 471
WhatsApp Image 2026-09-09 at 23 02 31
WhatsApp Image 2026-09-09 at 23 01 47

Using LLM to quickly evaluate alternatives, the common part with the PULS in this case could be a little class for the pwm cycle capture.

Please tell me your preferences about the implementation if you are interested in it. 
I will update the PR anyways after testing it on a real engine.

P.S Would have loved posting a pic of a FOME ECU into an Aston Martin :D

gdiciocco:
Verified OK on running engine.

image

PR #810, opened 2026-09-14
add honda J 6-2 cam trigger

jcor25:
## What

Adds a trigger decoder for the Honda J-series V6 intake cam wheel: six evenly spaced tooth slots with two of them missing, leaving four teeth — three normal gaps and one three times as wide per cam revolution.

It is a cam-only pattern, selected through the new `VVT_HONDA_J_6_2` ("Honda J 6-2") VVT mode rather than as a primary trigger, so it follows the same shape as the recent K24Z work:

- `TT_HONDA_J_CAM_6_2 = 79` with an `INVALID` placeholder in `trigger_type_e_enum`, matching how `TT_VVT_MAZDA_L`, `TT_VVT_MIATA_NA` and `TT_HONDA_K24Z_CAM_3` are handled — the user-facing name lives in `vvt_mode_e_enum`
- `VVT_HONDA_J_6_2 = 21`, wired through `getVvtTriggerType()` and the `adjustCrankPhase()` group that calls `syncAndReport(crankDivider, 0)`
- configured inline via `initializeSkippedToothTrigger(this, 6, 2, ...)`, consistent with the other generic skipped-tooth wheels in `initializeTriggerWaveform()`

Sync gap windows are 2.20–3.80 for the wide gap and 0.60–1.40 for the normal ones.

## Validation, and one caveat worth stating

Developed against a car with this cam, using tooth logger captures and datalogs. The original narrower windows (2.40–3.60 / 0.75–1.25) produced intermittent `CUSTOM_CAM_TOO_MANY_TEETH` warnings during cranking: the last normal gap ratio before the wide gap repeatedly measured 1.25–1.31, just over the limit, as cranking speed drifted. Every sync error in that log correlated with a ratio above 1.25. With the wider windows the observed ratios (~0.92–1.29 and ~3.13–3.30) sit comfortably inside, and a later 32-second log showed zero VVT sync errors.

**The tooth angles are modelled as evenly spaced 6-2, and that part is not fully verified.** In a tooth log the four gaps measured roughly 101°/122°/394°/104° of crank rotation against the modelled 120/120/120/360. Sync is reliable regardless, because the decoder works on gap *ratios* — but if the wheel really is unevenly spaced, the residual angle error would show up as error in reported VVT position. That capture was at ~228 rpm cranking, where within-cycle speed ripple is significant, so I can't currently separate real tooth spacing from speed variation. Verification at a steady running RPM is still outstanding.

If you'd rather this waited for that measurement, I'm happy to hold it. I've raised it explicitly rather than let it be discovered later.

## Testing

- New `unit_tests/tests/trigger/test_honda_j_cam.cpp`, modelled on `test_honda_k24z_cam.cpp`: feeds the gap sequence starting partway into the pattern, asserts sync lands on the tooth following the wide gap, that the index tracks, and that no warnings are raised.
- `cd unit_tests && make -j12 && ./build/fome_test` — 717 tests pass
- `cd firmware && make clean && make -j12` — clean from-scratch build passes
- `./format.sh check` — clean

`unit_tests/triggers.txt` is regenerated and included, since the auto-configs job doesn't maintain it. The enum-to-string headers, the TS inis and the Java enum are deliberately left out — the build regenerates them from the enums, and the auto-configs job commits them after merge.


mck1117:
do we have a photo of this wheel to confirm whether it's 60-60-60-180 or something else?

jcor25:
images (5)

mck1117:
Huh, looks like it really is evenly spaced, so 120 seems fine to me. The capture was probably wobbly as it was at cranking speed, so you're getting crank speed variation as each cylinder goes by.

PR #812, opened 2026-09-18
core8: disable internal pulls on VR/Hall trigger inputs

gdiciocco:
Core8 trigger inputs PE2–PE5 have external 5 V pull-ups. This change disables the STM32 internal pulls on those pins during early GPIO initialization, including in the bootloader, while preserving the defaults for all other pins.

Commit: `50f9da97fcc03bebb0536cf6e1d62b7d5112214b`


gdiciocco:
On my setup, with a VR crank trigger and a Hall cam trigger, the trigger frequently counted more than the expected 34 teeth on a 36-2 wheel. The engine could not be started. I disabled injection and ignition for diagnosis and checked the crank signal with an oscilloscope; it was clean both before and after the conditioner.

Further investigation led me to the internal pull configuration on the Core8 trigger inputs. The change in this PR resolved the extra tooth counts in my tests.

mck1117:
If the inputs on this hardware are unbuffered and pulled up to 5v, then this is a hardware design error. If they are buffered, then they aren't actually pulled up according to the STM32.

What is the actual hardware going on here?

gdiciocco:
> If the inputs on this hardware are unbuffered and pulled up to 5v, then this is a hardware design error. If they are buffered, then they aren't actually pulled up according to the STM32.
> 
> 
> 
> What is the actual hardware going on here?



FOME initializes PE2–PE5 with internal pull-downs. These pins are the four Core8 VR/Hall inputs and are STM32F407 FT pins.

The intent of this PR is to disable those internal pull-downs, not to change the external pull-up configuration. If the final Core8 conditioner/buffer drives the MCU inputs above VDD + 0.3 V, the STM32F407 datasheet requires the internal pull-up/down resistors to be disabled.

I agree the wording “PE2–PE5 have external 5 V pull-ups” may be misleading if the pull-up is before the buffer.

core8 schematics:

https://oshwlab.com/b.brazdziunas/scg-ecu_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy_copy

mck1117:
> core8 schematics:

core8 has functionally the same digital input circuitry as proteus/polygonus/atlas/etc, so this doesn't do anything. The pullup is before the buffer, so this *can't* do anything, as the STM32 pin is always hard driven high or low by the buffer.

gdiciocco:
Thanks for reviewing this. I think the key point is the distinction between the general digital inputs and the VR/Hall trigger inputs.

In the linked **v2.3 schematic**, the crank input I use follows **U5/MAX9926 COUT2 → VR4_CPU → PE5**, with an external pull-up path to 5 V through R105 and R123. The 74HC2G17 buffers appear on the separate `OUT_Dx` inputs.

The MAX9926’s **COUT1/COUT2 outputs are open-drain**, as specified on page 7 of the [[MAX9926 datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/MAX9924-MAX9927.pdf#page=7)](https://www.analog.com/media/en/technical-documentation/data-sheets/MAX9924-MAX9927.pdf#page=7). They actively drive low, but the high level depends on the external pull-up. Consequently, an internal pull-down on the STM32 can affect that node.

On the **STM32F407ZGT6**, PE2–PE5 are **FT (5 V tolerant) pins**. ST explicitly requires the internal pull-up and pull-down resistors to be disabled when the input voltage exceeds **VDD + 0.3 V**; see Table 48, note 5, page 117 of the [[STM32F407 datasheet](https://www.st.com/resource/en/datasheet/stm32f407vg.pdf#page=117)](https://www.st.com/resource/en/datasheet/stm32f407vg.pdf#page=117). This is the configuration requirement the patch addresses.

Empirically, on my setup, **this change allowed the engine to run with FOME for the first time**. At the point of turn I were debugging trigger issues with the oscilloscope connected and injection/ignitions disabled. The only difference was the patch. That is a meaningful improvement, but I cannot claim that the pull configuration was the sole cause: contributing factors or interactions with the rest of the hardware remain possible. I will take additional measurements on my board 5v stability.

mck1117:
does 92346bb3cfdd9168b9059176a83bec50a32d9936 fix your issue?

gdiciocco:
**Tested on real engine, fixes the issue**, no difference in behaviour from my excessively-specific patch.

Went a bit worried about other users of EXTI pins (auxiliaries mostly), but as you said shouldn't be causing issues if the hardware is correctly biased, and even rusefi applied a similar fix in 2024:

https://github.com/rusefi/rusefi/issues/6872
https://github.com/rusefi/rusefi/issues/2282

Problem solved and the fix is technically compliant, I'm however still trying to understand why such a low erosion of margin (theoretical 4.92V out of 5V) caused by the pull-down caused such a mayhem. I verified 5v stability with the oscilloscope with the engine running and it is fine, level appropriate and few mV of noise, didn't had time to go deeper on the signal to the pin. Will investigate further this weekend.

PR #813, opened 2026-09-18
Count dropped EXTI queue events

gdiciocco:
A full EXTI queue currently drops events without updating `extiOverflowCount`. Report a failed queue insertion and increment the counter at the point of loss. Saturate its 8-bit value so repeated losses cannot wrap it back to zero.

Validation: `git diff --check` passed. Firmware build was not run because `make` is unavailable in this Windows environment.


firmware/hw_layer/digital_input/digital_input_exti.cpp
```diff
@@ -163,7 +164,7 @@ class ExtiQueue {
 
 static CCM_OPTIONAL ExtiQueue<ExtiQueueEntry, 32> queue;
 
-static uint8_t overflowCounter = 0;
+static volatile uint8_t overflowCounter = 0;
```
mck1117: I don't think we actually need volatile here

gdiciocco: Yup, superfluous

mck1117:
looks like you've merge conflicted yourself with the other PR

gdiciocco:
darn, fixed, sorry

PR #814, opened 2026-09-18
Proposal: capture trigger pin level in the EXTI ISR

gdiciocco:
EXTI currently queues the interrupt timestamp, while the crank and cam callbacks read the pin later. If the level changes in between, an event may pair an earlier timestamp with a later level.

This proposes capturing the level when the event is queued and passing it to the crank and cam callbacks. Other EXTI users retain their existing interface. This is an optional improvement for upstream consideration.


mck1117:
why bother doing `withLevel` vs not? why not just capture it all the time? this code is very hot, so it's probably worth avoiding the branch.

gdiciocco:
Good point — I agree. I updated the PR so every EXTI event captures the pin level in the fast ISR and every callback receives it.

This removes `CallbackWithLevel`, the separate registration API, and the conditional branches from both the fast path and the deferred dispatcher. Callbacks that do not need the level simply ignore it; the trigger, flex-fuel, and Hella oil-level callbacks now consume the captured value.

I also benchmarked the alternatives on the actual Core8 STM32F4 at 168 MHz using `DWT->CYCCNT`.

**Conditional vs unconditional capture**

- conditional: 90,126 cycles / 4,096 events = 22.003 cycles/event
- unconditional: 77,837 cycles / 4,096 events = 19.003 cycles/event
- unconditional capture saves exactly 3 cycles/event

**Compared with the pre-patch deferred read**

- deferred edge-to-callback median: 348 cycles / 2.071 µs
- always-captured edge-to-callback median: 377 cycles / 2.244 µs
- difference: +29 cycles / 173 ns end-to-end

That full-path measurement used the earlier dual-callback implementation; the updated code also removes its deferred dispatcher branch.

A focused ISR measurement showed 178 cycles for the deferred implementation and 195 cycles for conditional capture. This means the maximum-priority ISR was extended by 17 cycles, or 101 ns. Unconditional capture removes 3 cycles from that path.

The pin itself was sampled after:

- deferred read: 2.071 µs median, 2.476 µs stable maximum
- ISR capture: 0.774 µs median

A simultaneous second EXTI added about 140 cycles / 0.83 µs to the deferred sampling delay.

For a nominal 36-2 wheel using both edges, the shortest normal edge interval is 83.3 µs at 10,000 rpm. RPM alone would only approach the deferred sampling window above roughly 336,000–402,000 rpm. Normal ECU interrupt load also does not significantly extend this window because the EXTI handoff runs at priority 3, above scheduling, ADC, UART, CAN, and USB.

This makes short additional transitions or noise within the 2–3 µs deferred window the likely failure mode, rather than normal tooth frequency. Capturing in the fast ISR reduces that vulnerable window to about 0.77 µs.

The timestamp is still taken before the GPIO read, so the current event timestamp is unchanged. The additional worst-case delay imposed on another event is about 101 ns with the conditional version, less another 3 cycles with unconditional capture. At 10,000 rpm, 101 ns corresponds to about 0.0061 crank degrees.

Builds completed successfully for Core8 and Hellen NB1.

PR #816, opened 2026-09-21
Visit only set output mask bits

gdiciocco:
`forEachSetBit` currently scans every position through the highest selected output, including unset bits. Visit the lowest set bit with a guarded `__builtin_ctz`, then clear it with `mask &= mask - 1`. Callback order remains ascending; the copied injector and ignition output contexts introduced in #551 and #553 are preserved.

The exhaustive unit test compares the callback sequence for all 65,536 masks. This branch starts directly from master `681f50477f` and contains only the helper change and its test.

Validation:

- Full unit suite: 717/717 passed; focused injection/ignition cases passed.
- Core8 and Hellen NB1 clean board-script builds passed; formatting and diff checks passed.
- Core8 hardware microbenchmark: STM32F4 at 168 MHz, GCC 11.3, normal `-O2`/LTO firmware flags. Both paths in one instrumented application call the same non-inlined observable callback, with no engine-output writes. Per case: warmup, nine alternating old/new trials of 4,096 walks, DWT timing with interrupts masked around the measured loop. Two complete runs agreed; all callback checksums matched.

Median cycles per walk (including the common callback and benchmark loop):

| Output mask | Before | After | Cycles saved |
| --- | ---: | ---: | ---: |
| `0x0000` | 6.01 | 7.01 | -1.00 |
| `0x0001` | 26.01 | 22.01 | 4.00 |
| `0x0008` | 44.02 | 22.01 | 22.00 |
| `0x0080` | 68.02 | 22.01 | 46.01 |
| `0x0800` | 92.02 | 22.01 | 70.01 |
| `0x000f` | 83.01 | 76.01 | 7.00 |
| `0x00ff` | 159.01 | 148.01 | 11.00 |
| `0xffff` | 311.01 | 292.01 | 19.00 |

These measurements isolate mask traversal with a controlled callback; they do not measure total fuel/spark callback latency or total ECU CPU usage. Empty-mask handling regresses by approximately one cycle in this harness. Benchmark instrumentation is excluded from the production diff. Other board configurations still require CI validation.


mck1117:
can we use the standard c++ `std::countr_zero` instead? or some other standard mechanism, rather than a non-portable gcc extension?

gdiciocco:
Agreed, `std::countr_zero` is the better choice here. The project already uses C++20, and the ARM toolchain supports it.

I’ve updated the implementation to use `<bit>` and `std::countr_zero`, preserving the existing guard, iteration order, and callback argument type. The linked ARM instructions are identical to the builtin version, and two ECU benchmark runs confirm equivalent performance.

All 717 unit tests and the Core8/Hellen NB1 builds pass.

mck1117:
this looks fine to me, ready to merge?

